PR #5, opened 2018-08-27
Improve the frontend

petr-muller:
Old argument structure and naming was chaotic. Now the generator has the
following options, consistently named and straightforward:

`--from-file`: Read ci-operator config from this file
`--from-dir`:  Walk and process all ci-operator configs under directory
`--from-release-repo`:
  Shortcut to `--from-dir=$GOPATH/src/github.com/openshift/release/ci-operator/config`

`--to-file`: Write generated jobs into this single file
`--to-dir`:  Write generated jobs into files under directory
`--to-release-repo`:
  Shortcut to `--to-dir=$GOPATH/src/github.com/openshift/release/ci-operator/jobs`

The --from and --to arguments can be combined arbitrarily.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -46,6 +46,32 @@ func bindOptions(flag *flag.FlagSet) *options {
 	return opt
 }
 
+func (o *options) process() error {
+	var err error
+
+	if o.fromReleaseRepo {
+		if o.fromDir, err = getReleaseRepoDir("ci-operator/config"); err != nil {
+			return fmt.Errorf("--from-release-repo error: %v", err)
+		}
+	}
+
+	if o.toReleaseRepo {
+		if o.toDir, err = getReleaseRepoDir("ci-operator/jobs"); err != nil {
+			return fmt.Errorf("--to-release-repo error: %v", err)
+		}
+	}
+
+	if (o.fromFile == "" && o.fromDir == "") || (o.fromFile != "" && o.fromDir != "") {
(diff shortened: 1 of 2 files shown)


cmd/ci-operator-prowgen/main.go
```diff
@@ -393,26 +411,54 @@ func main() {
...
-			if err != nil {
-				fmt.Fprintf(os.Stderr, "failed to write the job configuration (%v)\n", err)
+		} else if len(opt.toDir) > 0 {
+			if err := writeJobsIntoComponentDirectory(opt.toDir, org, repo, jobConfig); err != nil {
+				fmt.Fprintf(os.Stderr, "failed to write jobs to '%s' (%v)\n", opt.toDir, err)
 				os.Exit(1)
 			}
-
+		} else {
+			fmt.Fprintf(os.Stderr, "ci-operator-prowgen needs at least one of `--to-{file,dir,release-repo}` options\n")
```
stevekuznetsov: In a lot of our command-line tools we split up the input validation logic and execution logic so that when executing we can assume validity of input. Do you think that would potentially make this logic cleaner?

petr-muller: Good point. Yeah, I knew this was borderline when I committed it. I will invest a bit more work to simplify it. But in the end there will always be four slightly different cases in the flow, as the directory and file processing are quite different.

petr-muller: @stevekuznetsov PTAL now

PR #7, opened 2018-08-30
Label images postsubmit if images are promoted to 'openshift' 

petr-muller:
If the images built & promoted by the `--target=[images]` postsubmit are
promoted into the `openshift` namespace, the postsubmit job should have
the 'artifacts: images' label.

PR #8, opened 2018-08-30
Document what do we generate on what input

petr-muller:
Now when we have generation logic that is not absolutely trivial (especially with https://github.com/openshift/ci-operator-prowgen/pull/7), we should better document what do we want to generate on what input.

### DIFF
--- a/GENERATOR.md
+++ b/GENERATOR.md
@@ -1,0 +1,68 @@
+# Generated Prow jobs
+
+## Presubmits
+
+### Tests
+
+For each
+[test](https://github.com/openshift/ci-operator/blob/master/CONFIGURATION.md#tests)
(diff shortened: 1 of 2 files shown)


GENERATOR.md
```diff
@@ -0,0 +1,119 @@
...
+## Presubmits
+
+### Tests
+
+For each
+[test](https://github.com/openshift/ci-operator/blob/master/CONFIGURATION.md#tests)
+specified in the configuration file, generate one presubmit running ci-operator
+to build this test target.
+
+```yaml
```
stevekuznetsov, line 12: The whole YAML is not really useful. Maybe outline the:

 - job name
 - context name
 - rerun command

petr-muller:
I have shortened the YAML snippets to just the relevant parts.

PR #9, opened 2018-08-31
Change the ci-operator config format to YAML

petr-muller:
*this should be merged only after https://github.com/openshift/ci-operator/pull/121 is merged*

/hold

Parse ci-operator config files as YAML instead of JSON. Generated jobs do not change, except for the `ConfigMap` part, where we generate the filename to always  be a `.yaml`, so we won't  be able to generate jobs for `.json` files in the directory-walking mode. (I need to fix this)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ component?
 
 **Answer:**
 1. Get a working copy of [openshift/release](https://github.com/openshift/release) (we’ll shorten path to it to `$RELEASE`)
-2. Create a [ci-operator configuration file](https://github.com/openshift/ci-operator/blob/master/ONBOARD.md#prepare-configuration-for-component-repo) under `$RELEASE/ci-operator/config`, following the `organization/component/branch.json` convention.
+2. Create a [ci-operator configuration file](https://github.com/openshift/ci-operator/blob/master/ONBOARD.md#prepare-configuration-for-component-repo) under `$RELEASE/ci-operator/config`, following the `organization/component/branch.yaml` convention.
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
> so we won't be able to generate jobs for .json files in the directory-walking mode. (I need to fix this)

I don't understand this -- care to elaborate?

stevekuznetsov (commented):
/lgtm

petr-muller:
> I don't understand this -- care to elaborate?

I will rectify the underlying problem, but to explain: When passed `--from-dir`, the generator walks the directory and searches for all `*.json` (before) / `*.yaml` (after) files there. It considers its filename, without extension, to be a branch name. Based on the branch name, it then generates the jobs, including this part:

```yaml
- name: CONFIG_SPEC
  valueFrom:
    configMapKeyRef:
      key: branch.yaml
```

The `key: branch.yaml` part should match the source filename, but the source filename is not stored anywhere, just the `branch` part is, and the `key: branch.yaml` is generated simply by `sprintf("%s.yaml")`. Therefore, in the `--from-dir` mode, it needs to ignore `*.json` files because it would still generate `key: <name>.yaml` for such files, which would be broken. If the generator wants to support legacy json files (or even YAML files with the `.yml` suffix), it will need to correctly propagate full filename to this field.

Which is what I will do in next 20 minutes :)

petr-muller:
Fixed now. The diff is slightly obfuscated by https://github.com/openshift/ci-operator-prowgen/pull/9/commits/6620518eef497f3aa2c6da06e1dde6b2eabfa69a which just collapses all those `string, string, string, string` to a struct. I can PR is separately if you want.

PR #11, opened 2018-09-06
Fix triggering regexp

petr-muller:
In plain YAML string literals, backlashes do not need to be escaped

### DIFF
--- a/GENERATOR.md
+++ b/GENERATOR.md
@@ -16,7 +16,7 @@ to build this test target.
     context: ci/prow/TEST
     rerun_command: /test TEST
     spec: <pod that runs `ci-operator --target=TEST`>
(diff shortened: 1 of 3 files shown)


petr-muller:
Side note: `trigger: ` might be improved a bit . The `| all` part is [probably not even needed](https://github.com/kubernetes/test-infra/blob/86bc3ae52feb6ba31e9dd2510aa28d496bca7526/prow/plugins/trigger/ic.go#L28). The current [upstream default value](https://github.com/kubernetes/test-infra/blob/183533ae6ba748b8b07f4af16a49797c1b4df6a5/prow/config/config.go#L1079) is probably more readable and allowing more things, like `\test unit lint`. I think I'll switch the generator to use that default value.

PR #14, opened 2018-09-19
update vendors

droslean:


stevekuznetsov:
@droslean is this the result of updating just ci-operator? 

https://golang.github.io/dep/docs/daily-dep.html#updating-dependencies

PR #19, opened 2018-09-21
Refactor generator to write at the directory level

stevekuznetsov:
We want to enforce that jobs are sharded by the branch and the job type,
which we can only do if we load and write at the directory level,
instead of the single file level. This patch reconfigured the generator
and the determinizer to operate at the directory level and to shard jobs
by branch and by job type.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @bbguimaraes @petr-muller

PR #21, opened 2018-09-24
Always pull the images for tests

stevekuznetsov:
We use rolling tags for images used for tests, so we need to ensure that
those are always pulled down before tests are run or rollouts to update
the images underlying tests are nondeterministic.

/assign @petr-muller @bbguimaraes


stevekuznetsov:
Integration will obviously fail until #19 merges

stevekuznetsov:
Implemented: https://github.com/openshift/release/pull/1650

stevekuznetsov:
Everything but the integration is passing, going to merge this guy

PR #22, opened 2018-09-25
Don't write old jobs if they're being written elsewhere

stevekuznetsov:

/cc @petr-muller


petr-muller:
:+1:

PR #25, opened 2018-10-02
change config map usage.

droslean:
This PR changes the configMap name and key configuration.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -75,11 +75,11 @@ func (o *options) process() error {
 // Generate a PodSpec that runs `ci-operator`, to be used in Presubmit/Postsubmit
 // Various pieces are derived from `org`, `repo`, `branch` and `target`.
 // `additionalArgs` are passed as additional arguments to `ci-operator`
-func generatePodSpec(org, repo, configFile, target string, additionalArgs ...string) *kubeapi.PodSpec {
+func generatePodSpec(configFile, target string, additionalArgs ...string) *kubeapi.PodSpec {
 	configMapKeyRef := kubeapi.EnvVarSource{
 		ConfigMapKeyRef: &kubeapi.ConfigMapKeySelector{
 			LocalObjectReference: kubeapi.LocalObjectReference{
-				Name: fmt.Sprintf("ci-operator-%s-%s", org, repo),
+				Name: "ci-operator-configs",
 			},
 			Key: configFile,
 		},
@@ -121,7 +121,7 @@ func generatePresubmitForTest(test testDescription, repoInfo *configFilePathElem
 		Context:      fmt.Sprintf("ci/prow/%s", test.Name),
 		Name:         name,
 		RerunCommand: fmt.Sprintf("/test %s", test.Name),
-		Spec:         generatePodSpec(repoInfo.org, repoInfo.repo, repoInfo.configFilename, test.Target, additionalArgs...),
+		Spec:         generatePodSpec(repoInfo.configFilename, test.Target, additionalArgs...),
 		Trigger:      fmt.Sprintf(`((?m)^/test( all| %s),?(\s+|$))`, test.Name),
 		UtilityConfig: prowconfig.UtilityConfig{
 			DecorationConfig: &prowkube.DecorationConfig{SkipCloning: true},
@@ -144,7 +144,7 @@ func generatePostsubmitForTest(
 		Agent:    "kubernetes",
 		Brancher: prowconfig.Brancher{Branches: []string{repoInfo.branch}},
 		Name:     name,
-		Spec:     generatePodSpec(repoInfo.org, repoInfo.repo, repoInfo.configFilename, test.Target, additionalArgs...),
+		Spec:     generatePodSpec(repoInfo.configFilename, test.Target, additionalArgs...),
 		Labels:   labels,
 		UtilityConfig: prowconfig.UtilityConfig{
 			DecorationConfig: &prowkube.DecorationConfig{SkipCloning: true},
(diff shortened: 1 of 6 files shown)


droslean:
/hold
Hold until changes in release repo.

cmd/ci-operator-prowgen/main.go
```diff
@@ -263,7 +263,8 @@ func extractRepoElementsFromPath(configFilePath string) (*configFilePathElements
 	}
 
 	fileName := filepath.Base(configFilePath)
-	branch := strings.TrimSuffix(fileName, filepath.Ext(configFilePath))
+	s := strings.Split(strings.TrimSuffix(fileName, filepath.Ext(configFilePath)), "-")
+	branch := s[len(s)-1]
```
petr-muller: Hmm this won't work when branch name contains the `-` separator, like our `release-X.Y` branches do. I think you'll need to `TrimPrefix` the `$org-$repo-` to get the proper branch name.

petr-muller (changes requested):

cmd/ci-operator-prowgen/main.go
petr-muller: That is actually what the `ci/prow/integration` test seem to be failing on - the branch name is extracted wrong.

cmd/ci-operator-prowgen/main.go
```diff
@@ -263,7 +264,13 @@ func extractRepoElementsFromPath(configFilePath string) (*configFilePathElements
...
 
 	fileName := filepath.Base(configFilePath)
-	branch := strings.TrimSuffix(fileName, filepath.Ext(configFilePath))
+	s := strings.Split(strings.TrimSuffix(fileName, filepath.Ext(configFilePath)), "-")
+
+	branch := s[len(s)-1]
+
+	if _, err := strconv.ParseFloat(branch, 64); err == nil {
+		branch = s[len(s)-2] + "-" + s[len(s)-1]
+	}
```
petr-muller: This will still break on branch names using `-` sign where the last part is not numeric. Like `openshift-master`, `whatever-testing` etc.

petr-muller (changes requested):

cmd/ci-operator-prowgen/main.go
```diff
@@ -263,9 +263,16 @@ func extractRepoElementsFromPath(configFilePath string) (*configFilePathElements
 	}
 
 	fileName := filepath.Base(configFilePath)
-	branch := strings.TrimSuffix(fileName, filepath.Ext(configFilePath))
+	s := strings.Split(strings.TrimSuffix(fileName, filepath.Ext(configFilePath)), "-")
 
-	return &configFilePathElements{org, repo, branch, fileName}, nil
+	var branchSlice []string
+	for _, v := range s {
+		if v != org || v != repo {
```
petr-muller: This is still not correct. As written, the condition will be always true unless `org == repo`, extracting whole filename (sans extension). I assume you wanted to put `&&` there, but then it breaks on branch names that contain either `org` or `repo`, like this:

`openshift-prometheus-operator-openshift-master` ([this is a real example](https://github.com/openshift/prometheus-operator/tree/openshift-master)) - this would be extracted as just `master`.

I really suggest doing something like:
```
basename := strings.TrimSuffix(fileName, filepath.Ext(configFilePath))
branch := string.TrimPrefix(basename, fmt.Sprintf("%s-%s-", org, repo))
```

It is entirely possible I am overlooking some reason why this would not work, though :)

petr-muller (changes requested):

cmd/ci-operator-prowgen/main.go
droslean: Maybe I missed something here, but my tests say the opposite. I think I tested all the cases and this code is correct.

petr-muller: My assumption is that we are renaming all ci-operator configuration files to have `$org-$repo-$branch.yaml` format. Under that assumption, a config for [openshift/prometheus-operator@openshift-master](https://github.com/openshift/prometheus-operator/tree/openshift-master) would be stored in this file:

```
ci-operator/config/openshift/prometheus-operator/prometheus-operator-openshift-master.yaml
```

As written, this code would extract branch name incorrectly as `openshift-prometheus-operator-openshift-master` (https://play.golang.org/p/8-mQb9z-IOe)

droslean: `ci-operator/config/openshift/prometheus-operator/master.yaml` renamed to `ci-operator/config/openshift/prometheus-operator/openshift-prometheus-operator-master.yaml`

droslean:
Let us keep it like this, and assume that the renaming was done in format `$org-$repo-$branch`. Let's try to validate this in other scripts. So +1 @petr-muller

petr-muller:
Cool. I like how the feature makes the code simpler!

petr-muller:
/lgtm
/hold

I'm holding this so we can sync the merge with making https://github.com/openshift/release/pull/1778 up-to-date

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -609,7 +603,7 @@ func TestFromCIOperatorConfigToProwYaml(t *testing.T) {
           valueFrom:
             configMapKeyRef:
               key: branch.yaml
```
stevekuznetsov, line 605: This is wrong, yeah?

test/integration/data/output/jobs/super/duper/super-duper-master-presubmits.yaml
```diff
@@ -103,7 +103,7 @@ presubmits:
           valueFrom:
             configMapKeyRef:
               key: master.yaml
```
stevekuznetsov, line 105: These are also wrong?

stevekuznetsov (commented):
Some tests need updating

cmd/ci-operator-prowgen/main_test.go
petr-muller: You mean wrong by not having `$org-$repo-$branch.yaml` format? Yeah. 

It will still extract the branch name "right" - it will attempt to remove a `super-duper-` prefix and `branch` will remain :)

petr-muller:
/lgtm cancel

Let's fix the tests @stevekuznetsov spotted...

cmd/ci-operator-prowgen/main_test.go
stevekuznetsov: Fun side-effect! Backwards compatiblity

PR #26, opened 2018-10-02
Sort fields in PodSpec for deterministic output

stevekuznetsov:

/assign @bbguimaraes


petr-muller:
We just need to be aware this breaks things if we for any reason need to have positional args.

stevekuznetsov:
True -- we'll keep that in mind if we ever are to use them

PR #28, opened 2018-10-02
Ensure all additional args are in the --key=val form

stevekuznetsov:

/assign @petr-muller 

We have a couple options:

 - enforce this in review only
 - use a panic like this
 - enforce this downstream of the spec generation somewhere by adding an optional `error` return to everything

What do you think is better?


petr-muller:
I think the panic is fine for now. I don't think it's worth the effort to put there a error-propagating structure just for this, and I think this gets easily overlooked in reviews especially when we ask people to run the determinizer and blindly trust the result.

stevekuznetsov:
@bbguimaraes since you're touching this in your PR, I will let this lapse so we can get the other one in without more rebase.

bbguimaraes:
@stevekuznetsov: I'm reusing this function partly because I got tired of rebasing, it's fine.

PR #34, opened 2018-10-10
Add new cluster profiles - aws-gluster and gcp-logging

vrutkovs:
Add new cluster types, requires https://github.com/openshift/ci-operator/pull/176

Now sure if I should bump up ci-operator dep manually or its gonna happen automagically


vrutkovs:
Will be resubmitted along with new test types

PR #36, opened 2018-10-12
Add new openshift-ansible test types and profiles in the generator

vrutkovs:
Verified that this can generate GCP logging, custom aws template and GCP major upgrade job

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -121,6 +121,14 @@ func generatePodSpecTemplate(org, repo, configFile, release string, test *cioper
 		template = "cluster-launch-src"
 		clusterProfile = conf.ClusterProfile
 		needsReleaseRpms = true
+	} else if conf := test.OpenshiftAnsibleCustomClusterTestConfiguration; conf != nil {
+		template = "cluster-launch-openshift-ansible"
+		clusterProfile = conf.ClusterProfile
+		needsReleaseRpms = true
+	} else if conf := test.OpenshiftAnsibleUpgradeClusterTestConfiguration; conf != nil {
+		template = "cluster-launch-e2e-upgrade"
+		clusterProfile = conf.ClusterProfile
+		needsReleaseRpms = true
 	} else if conf := test.OpenshiftInstallerClusterTestConfiguration; conf != nil {
 		template = "cluster-launch-installer-e2e"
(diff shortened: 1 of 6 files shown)


cmd/ci-operator-prowgen/main.go
```diff
@@ -112,7 +112,10 @@ func generatePodSpec(configFile, target string, additionalArgs ...string) *kubea
 func generatePodSpecTemplate(org, repo, configFile, release string, test *cioperatorapi.TestStepConfiguration, additionalArgs ...string) *kubeapi.PodSpec {
 	var template string
 	var clusterProfile cioperatorapi.ClusterProfile
+	var isUpgradeConfig bool
 	var needsReleaseRpms bool
+	var previousVersion string
```
vrutkovs: not entirely sure about those - should I copy config string here or just make `conf` visible later on?

cmd/ci-operator-prowgen/main.go
```diff
@@ -193,6 +207,14 @@ func generatePodSpecTemplate(org, repo, configFile, release string, test *cioper
...
 		})
 	}
+	if isUpgradeConfig {
+		container.Env = append(
+			container.Env,
+			kubeapi.EnvVar{Name: "PREVIOUS_ANSIBLE_VERSION", Value: previousVersion},
+			kubeapi.EnvVar{Name: "PREVIOUS_IMAGE_ANSIBLE", Value: fmt.Sprintf("docker.io/openshift/origin-ansible:v%s", previousVersion)},
+			kubeapi.EnvVar{Name: "PREVIOUS_RPM_DEPENDENCIES_REPO", Value: previousRPMDeps},
+			kubeapi.EnvVar{Name: "PREVIOUS_RPM_REPO", Value: fmt.Sprintf("https://rpms.svc.ci.openshift.org/openshift-origin-v%s/", previousVersion)})
+	}
```
bbguimaraes, line 215: As you said above, since these are specific to a single type of test, remove the local variables and check the field directly:

```go
if conf := test.OpenshiftAnsibleUpgradeClusterTestConfiguration; conf != nil {
    // use conf.PreviousVersion, etc.
}
```

bbguimaraes (changes requested):

cmd/ci-operator-prowgen/main.go
vrutkovs: Done

PR #37, opened 2018-10-12
When generating the postsubmit images job, only promote on exact matches

smarterclayton:
Because prowjob branches are regexps, the "master" branch matches
`.*master.*`. If a team has a working branch like `master-dockerless`
and merges to that branch, the promote images job will stomp the images
that the `master` branch promotes.

With this change, the image promotion job is set to exactly match the
requested branch, unless the branch looks like a regex (has characters
outside the normal branch name set). This prevents accidental image
promotion that occurs on feature branches.

Yesterday master jobs were broken for several hours because the master-dockerless
branch was pushed to and the promotion job promoted it to openshift/origin-v4.0,
overwriting the artifacts image with the contents of that branch (it is much older)
and causing older RPMs to not be served.

https://openshift-gce-devel.appspot.com/build/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-4.0/768

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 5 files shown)


smarterclayton:
@stevekuznetsov @bbguimaraes I can regenerate the jobs once this merges.

smarterclayton:
Looks like this breaks 

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_release/1919/pull-ci-openshift-release-master-ordered-prow-config/1502

and 

https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_release/1919/pull-ci-openshift-release-master-prow-config-semantics/692

The fact that branches are regexes needs to be taken into account in these jobs.

Nevermind, these will also be fixed when the PR lands (I tested locally)

smarterclayton:
Where does `validate-ci-operator-config-filename.py` come from in `registry.svc.ci.openshift.org/ci/python-validation`?

EDIT: nm, found it

smarterclayton:
Without this change merges to `release-3.1` would match `release-3.11` (not a problem now, but could very quickly become one).

bbguimaraes:
That's interesting, I don't think anyone was aware that the field is treated as a regular expression.

That said, do we want to support generating jobs with regular expressions in that field? Jobs already have to be branch-specific because of the ci-operator configuration.

The branch name is extracted from the configuration file name, which could have problematic characters (`.` in particular), but this could be solved by always anchoring and escaping.

@petr-muller @stevekuznetsov

stevekuznetsov:
Yeah, I'm thinking similarly ... @smarterclayton you asked for the branch regex, but we've never used it. Furthermore, copying jobs from branch A to B -- using the `ci-operator` and generator, is trivial. Do we need the regex? Should I just remove that feature upstream?

smarterclayton:
a key goal of 4.0 release/CI is to make it easy to carry CI jobs to short
lived branches so that we can CI test feature branches and embargo fixes
easily.  Ben was using the regex (that’s where the bug came from - it was
deliberate that presubmit jobs would run on that branch exactly like master
jobs).

It’s possible that upstream should have a less surprising rule than “always
a regex”, but then we’re more into globs than regex.  Either way, we should
have a design for “how do we have multiple long running bug/feature
branches that get CI and keep the amount of config changes low”.


On Oct 15, 2018, at 6:02 AM, Steve Kuznetsov <notifications@github.com>
wrote:

Yeah, I'm thinking similarly ... @smarterclayton
<https://github.com/smarterclayton> you asked for the branch regex, but
we've never used it. Furthermore, copying jobs from branch A to B -- using
the ci-operator and generator, is trivial. Do we need the regex? Should I
just remove that feature upstream?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/openshift/ci-operator-prowgen/pull/37#issuecomment-429784350>,
or mute the thread

PR #38, opened 2018-10-15
Be more intelligent about warning for long names

stevekuznetsov:
When a repo has such a long prefix that there's no choice but to run
over the soft limit on job name, there's no point in warning about it.
However, when a user has the option to make a reasonable name for their
job and not run over the limit, but they don't, we should warn them.

/assign @petr-muller


petr-muller:
Pretty cool.

/lgtm

PR #42, opened 2018-11-06
Openshift installer src template

ecordell:
This adds support for the cluster-launch-installer-src template, which runs a test command in the `src` image after installing a cluster with openshift installer.


ecordell:
Used an older version of dep - I'll update with a fixed lock file

PR #43, opened 2018-11-10
Support `variant` on ci-operator config definitions

smarterclayton:
Part of https://github.com/openshift/release/pull/2102

When a definition has a trailing segment separated from branch by an
underscore, treat it as a separate type of job:

openshift-builder-master.yaml
openshift-builder-master_rhel.yaml

becomes job definitions:

pull-openshift-builder-master-rhel-images
pull-openshift-builder-master-rhel-e2e-aws
etc

Will be used to allow us to add RHEL variants for our core definitions,
gradually move things over to them, and then switch.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -238,6 +238,9 @@ type testDescription struct {
 // Generate a Presubmit job for the given parameters
 func generatePresubmitForTest(name string, repoInfo *configFilePathElements, podSpec *kubeapi.PodSpec) *prowconfig.Presubmit {
 	jobPrefix := fmt.Sprintf("pull-ci-%s-%s-%s-", repoInfo.org, repoInfo.repo, repoInfo.branch)
+	if len(repoInfo.variant) > 0 {
+		name = fmt.Sprintf("%s-%s", repoInfo.variant, name)
+	}
 	jobName := fmt.Sprintf("%s%s", jobPrefix, name)
 	if len(jobName) > 63 && len(jobPrefix) < 53 {
 		// warn if the prefix gives people enough space to choose names and they've chosen something long
@@ -268,6 +271,9 @@ func generatePostsubmitForTest(
 	podSpec *kubeapi.PodSpec) *prowconfig.Postsubmit {
 	branchName := jc.MakeRegexFilenameLabel(repoInfo.branch)
 	jobPrefix := fmt.Sprintf("branch-ci-%s-%s-%s-", repoInfo.org, repoInfo.repo, branchName)
+	if len(repoInfo.variant) > 0 {
+		name = fmt.Sprintf("%s-%s", repoInfo.variant, name)
+	}
 	jobName := fmt.Sprintf("%s%s", jobPrefix, name)
 	if len(jobName) > 63 && len(jobPrefix) < 53 {
 		// warn if the prefix gives people enough space to choose names and they've chosen something long
@@ -397,6 +403,7 @@ type configFilePathElements struct {
 	org            string
 	repo           string
 	branch         string
+	variant        string
 	configFilename string
 }
 
@@ -420,7 +427,13 @@ func extractRepoElementsFromPath(configFilePath string) (*configFilePathElements
 	s := strings.TrimSuffix(fileName, filepath.Ext(configFilePath))
 	branch := strings.TrimPrefix(s, fmt.Sprintf("%s-%s-", org, repo))
 
-	return &configFilePathElements{org, repo, branch, fileName}, nil
+	var variant string
+	if i := strings.LastIndex(branch, "__"); i != -1 {
+		variant = branch[i+2:]
+		branch = branch[:i]
+	}
+
+	return &configFilePathElements{org, repo, branch, variant, fileName}, nil
 }
 
 func generateProwJobsFromConfigFile(configFilePath string) (*prowconfig.JobConfig, *configFilePathElements, error) {


smarterclayton:
@stevekuznetsov @petr-muller @bbguimaraes @droslean need some opinions here.  We want to build images slightly differently from the same branch - my RHEL example PR shows how this would be used.  We would gradually add more images to the RHEL side, then flip the e2es to be part of the RHEL variant, and gradually scale down the non-RHEL set.

petr-muller:
The obvious pitfall is that `_` is a valid git branch name character so whenever anyone would try to set up ci-operator jobs for branch containing `_` it could lead to quite surprising bugs. Currently we don't have any but I imagine someone will create one as soon as we would merge this. Maybe we can use `..` as an optional separator - it's ugly but it's illegal as a branch name.

Otherwise, I don't see any clearly better approach for the solution than this. Re-engineering ci-operator config to allow specifying an intent to build several sets of images from the same source is probably not worth the effort.

What might be possible is to modify the generator to generate modified jobs straight from the original ci-operator config file. Such jobs would, of course, need the modified config files at runtime, so they would need to derive them from the original ones - that seems to be possible. the differences between `_rhel` and original  config seems to be stripping all content not related to building images (easy), using a different namespace (easy) and sometimes different Dockerfile paths (some pitfalls). An advantage would be that we wold not need to maintain duplicate ciop configs.

But if this is a temporary measure until we migrate images to be RHEL-based, then I think the approach in this PR is totally fine.

bbguimaraes:
My primary concern is the maintenance cost of forking the files. Can't the builds in the existing configuration files be changed instead of keeping two different files at the same time? If that is not possible, this seems like the least disruptive solution.

Also, the images in ci-operator namespaces are publicly accessible, if that is a concern.

> We would gradually add more images to the RHEL side, then flip the e2es to be part of the RHEL variant, and gradually scale down the non-RHEL set.

We can keep the tests on both files until they are green on the RHEL side.

stevekuznetsov:
If we add a field to the ci-operator config YAML for a name prefix on jobs, it will be silently ignored by ci-operator but could be used by the generator here. Might be preferable to flattening more structured data into strings before parsing it back out ...

petr-muller:
> If we add a field to the ci-operator config YAML for a name prefix on jobs, it will be silently ignored by ci-operator but could be used by the generator here. Might be preferable to flattening more structured data into strings before parsing it back out ...

The filenames would still need to be _somehow_ different, as we use them as keys to a single `ConfigMap`...

stevekuznetsov:
(shit)

droslean:
@smarterclayton @stevekuznetsov @petr-muller @bbguimaraes 
I would prefer to have this in the same file, and I tried to find a good way to support it.  What I am suggesting is to support `pipelines` in the same configuration file. 
Example:

```yaml
pipelines:
  rhel:
    base_images:
      base:
        name: '4.0'
        namespace: ocp
        tag: base
    build_root:
      image_stream_tag:
        name: builder
        namespace: ocp
        tag: golang-1.10
    canonical_go_repository: github.com/openshift/builder
    images:
    - dockerfile_path: Dockerfile.rhel7
      from: base
      to: docker-builder
    resources:
      '*':
        limits:
          cpu: '2'
          memory: 4Gi
        requests:
          cpu: 100m
          memory: 200Mi
    tag_specification:
      name: '4.0'
      namespace: ocp

  default:
    tag_specification:
      cluster: https://api.ci.openshift.org
      name: origin-v4.0
      namespace: openshift
    base_images:
      base:
        cluster: https://api.ci.openshift.org
        name: origin-v4.0
        namespace: openshift
        tag: base
    build_root:
      image_stream_tag:
        cluster: https://api.ci.openshift.org
        name: release
        namespace: openshift
        tag: golang-1.10
    canonical_go_repository: github.com/openshift/builder
    images:
    - from: base
      to: docker-builder
    tests:
    - as: unit
      commands: hack/test.sh
      container:
        from: src
    - as: verify
      commands: hack/verify.sh
      container:
        from: src
    resources:
      '*':
        limits:
          cpu: '2'
          memory: 4Gi
        requests:
          cpu: 100m
          memory: 200Mi
```
Then we can specify the pipeline with a flag `--pipeline=rhel`. Of course we don't want to change all our existing configuration files, **so if the pipeline doesn't exists, we will continue to treat the configuration file as before.**

The implementation should be easy and it won't break anything.

stevekuznetsov:
@droslean that's not a backwards compatible change to the configuration file and would require migration for all the others. I think the filename change is less intrusive.

smarterclayton:
Do we want to sync on this next week?  I can join a scrum meeting or something and we can discuss at the end.

stevekuznetsov:
I don't have any other comments here, I think this approach is reasonable for what we're trying to achieve. @droslean @petr-muller @bbguimaraes thoughts?

petr-muller:
If we need this only temporarily I'm also leaning towards it being fine. Please consider using `..` as a separator though - it's ugly but ugly is better than funny side effects. And ugly does not hurt if we only do this temporarily.

I'd like to support what Bruno asked though - why do we *need* the two versions? Why don't we just change the existing configs to RHEL pipeline?

droslean:
I agree that if we don't care to have both pipelines in the future, this solution seems logical. Also, I am not concern about the `_` delimiter. If that happens we can detect it, it's not so critical. 
It is also worth considering what @bbguimaraes asked.

smarterclayton:
`..` as a separator?  That's super unusual.  How about two underscores?  That would be something I would prevent people creating git repos on, and has the extra visual spacing to make it noticeable.  

We have to make a transition slowly.  There will be two pipelines while we test the various different types of stacks.  If we Big Bang this, we're likely cause widespread disruption and delay the release significantly (because we have so many projects, each one has to change gradually).  Some images don't have the infra to switch to this yet (logging / console need RPMs mirrored out of the firewall).  So it has to be a measured change, but the end state should be to converge back to one.  We'll probably start by generating all the golang-no_dependencies projects.  Then the golang-some_dependencies.  Then the simple_runtime-no_dependencies.  Then the simple_runtime-complex_dependencies (logging).

I will note that we will probably have CI on alternate architectures at some point (arm) or alternate versions of RHEL (8 vs 7 etc).  Building has to be multiplexed somehow.

smarterclayton:
Changed to `__`

petr-muller:
Seems we had a "let's just get rid of this" consensus on slack.

/lgtm
/approve
/meow

petr-muller:
Very appropriate.

smarterclayton:
that cat is adorable

PR #44, opened 2018-11-13
Use openshift/openshift-ansible subfolder for templates

vrutkovs:
This is a precondition to rework templates location so that it would easier to support 3.1x and 4.0 branches


vrutkovs:
That doesn't seem to be required

PR #47, opened 2018-12-03
Give all jobs a variant label that need one

smarterclayton:
Labels should be used to carry categorization info to jobs, and variant
is an excellent one (like artifacts).  Allows validation in https://github.com/openshift/release/pull/2102 to correctly identify variants.

### DIFF
--- a/cmd/ci-operator-prowgen/main_test.go
+++ b/cmd/ci-operator-prowgen/main_test.go
@@ -734,6 +734,7 @@ func TestFromCIOperatorConfigToProwYaml(t *testing.T) {
 		org                        string
 		component                  string
 		branch                     string
+		variant                    string
 		configYAML                 []byte
 		prowOldPresubmitYAML       []byte
 		prowOldPostsubmitYAML      []byte
@@ -887,6 +888,7 @@ tests:
 			org:       "super",
 			component: "duper",
 			branch:    "branch",
+			variant:   "rhel",
 			configYAML: []byte(`base_images:
   base:
     cluster: https://api.ci.openshift.org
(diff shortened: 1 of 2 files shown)


cmd/ci-operator-prowgen/main_test.go
```diff
@@ -388,10 +397,13 @@ func TestGeneratePostSubmitForTest(t *testing.T) {
 			treatBranchesAsExplicit: true,
 
 			expected: &prowconfig.Postsubmit{
-				Agent:    "kubernetes",
+				Agent: "kubernetes",
+				Labels: map[string]string{
+					"artifacts":                        "images",
+					"ci-operator.openshift.io/variant": "",
```
stevekuznetsov: Add the label values as a constant please

test/integration/data/output/jobs/super/duper/super-duper-release-3.11-postsubmits.yaml
```diff
@@ -6,6 +6,7 @@ postsubmits:
     decorate: true
     labels:
       artifacts: images
+      ci-operator.openshift.io/variant: ""
```
stevekuznetsov: Can we treat a missing label the same way as empty? I don't want to see this add a linearly scaling number of lines to our config ...

smarterclayton: we could, but the empty value is a lot nicer for people to integrate with.  it's up to you.  I think about all the places where someone wants to find and replace etc.

petr-muller:
/retest
how about now?

petr-muller:
/retest
and now?

petr-muller:
/retest
pretty please?

test/integration/data/output/jobs/super/duper/super-duper-release-3.11-postsubmits.yaml
stevekuznetsov: I don't see how it's easier to integrate -- if you're opting in you're adding a line anyway right?

smarterclayton: I can remove it, makes the change smaller

smarterclayton:
Ok, comments addressed.  Label is optional, has a constant.

petr-muller (approved):

PR #48, opened 2018-12-17
Only promote to openshift:origin-v4.0 from master branches

petr-muller:
Adding a sanity check to avoid promoting to important tags from feature
branches, when using the default promotion targets.

Avoiding gotchas like the one attempted to be fixed in https://github.com/openshift/release/pull/2413

/cc @stevekuznetsov @smarterclayton

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -340,6 +340,18 @@ func extractPromotionName(configSpec *cioperatorapi.ReleaseBuildConfiguration) s
 	return ""
 }
 
+func shouldBePromoted(branch, namespace, name string) bool {
+	if namespace == "openshift" {
+		if name == "origin-v4.0" {
+			return branch == "master"
+		}
+
+		// TODO: release branches?
+	}
+
+	return true
+}
+
 // Given a ci-operator configuration file and basic information about what
 // should be tested, generate a following JobConfig:
 //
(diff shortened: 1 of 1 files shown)


cmd/ci-operator-prowgen/main.go
```diff
@@ -386,7 +400,11 @@ func generateJobs(
 		}
 
 		presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest("images", repoInfo, generatePodSpec(repoInfo.configFilename, "[images]", additionalPresubmitArgs...)))
-		postsubmits[orgrepo] = append(postsubmits[orgrepo], *generatePostsubmitForTest("images", repoInfo, true, labels, generatePodSpec(repoInfo.configFilename, "[images]", additionalPostsubmitArgs...)))
+
+		// If we have and explicit promotion config, let's respect that. Otherwise, validate if the branch matches promotion target
+		if configSpec.PromotionConfiguration != nil || shouldBePromoted(repoInfo.branch, promotionNamespace, promotionName) {
```
smarterclayton, line 405: I would go so far as to say this should be `isPromotionBranch` and allow only `release-X.Y` and `master` to be promotion branches.

smarterclayton: Although if we want to say that the correct place to do it would be in the ci-operator config, not here.  So that inclines me to say "let's not do anything here yet" and instead make the switch in ci-operator config to stop defaulting.

Also, we really need to ensure that promotion can't happen from non-exact branches.  A job that runs on `release-3.11-X` shouldn't promote to `3.11`.  We have that in the generators, but I'm actually not sure we have that protection in ci-operator.

petr-muller: Yeah, probably even better

stevekuznetsov:
Seems like the right low-cost fix
/lgtm

PR #49, opened 2018-12-19
Add a capability to execute rehearsal jobs

petr-muller:
The `rehearse` binary takes a PR on a repository holding Prow job
configuration (openshift/release by default) and submits ephemeral
"rehearsal" Prow jobs. Rehearsal jobs are jobs configured for a
different OpenShift repo but executed on openshift/release repo PR to
give Prow job configuration authors early feedback on how their changes
will affect "production" Prow job executions.

Current version of rehearse mocks the "which jobs need be be executed"
code and only focuses on the actual rehearsal job creation and
submission.

If you manage to compile it, it can be used together with my [test PR](https://github.com/openshift/release/pull/2367), like this (you need to pass the PR2367 version of the files to `rehearse`):

```
$ ./rehearse --github-token-path=oauth --config-path=.../openshift/release/cluster/ci/config/prow/config.yaml --job-config-path=.../openshift/release/ci-operator/jobs/openshift/ci-operator/ --pr 2367
``` 
Currently, the tool is totally dumb, it only makes the modified and re-targeted Presubmit prowjobs and creates them in the cluster. It does not yet create any configmaps nor other resources for the created jobs to consume (that's my next step).

### WIP items

- [x] Finish dependency resolution
I have worked on the code in a standalone repo and tried to add the deps to `Gopkg.toml`, but my `dep ensure` fails on something that looks like the `test-infra` dependency path needs new-ish kubernetes version and the `ci-operator` path needs something old.
- [x] Optimize the clientset creation
I noticed I have a leftover code in that creates a ProwJob clientset for each job submitted, which should only be done once


petr-muller:
Oh, and there is a buildable standalone version (without dep conflicts) in my repo here: https://github.com/petr-muller/pj-rehearse
The code should be the same ATM.

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/pjutil"
+
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+func getJobsToExecute(ca *config.Agent) config.JobConfig {
+	var youHaveOneJob config.Presubmit
```
stevekuznetsov: lol nice

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+
+func makeRehearsalPresubmit(source *config.Presubmit, repo string, prNumber int) (*config.Presubmit, error) {
+	var rehearsal config.Presubmit
+	deepcopy.Copy(&rehearsal, source)
+
+	rehearsal.Name = fmt.Sprintf("rehearse-%d-%s", prNumber, source.Name)
+	rehearsal.Context = fmt.Sprintf("ci/rehearse/%s/%s", repo, strings.TrimPrefix(source.Context, "ci/prow/"))
+
+	if len(source.Spec.Containers) != 1 {
+		return nil, fmt.Errorf("Cannot rehearse jobs with more than 1 container in Spec")
```
stevekuznetsov: ProwJobs with more than one container are entirely invalid

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	container := source.Spec.Containers[0]
+
+	if len(container.Command) != 1 || container.Command[0] != "ci-operator" {
+		return nil, fmt.Errorf("Cannot rehearse jobs that have Command different from simple 'ci-operator'")
```
stevekuznetsov: Why?

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	for _, arg := range container.Args {
+		if strings.HasPrefix(arg, "--git-ref") || strings.HasPrefix(arg, "-git-ref") {
+			return nil, fmt.Errorf("Cannot rehearse jobs that call ci-operator with '--git-ref' arg")
```
stevekuznetsov: How many of these exist? That is antithetical to how `ci-operator` was supposed to be used

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+
+	if len(source.Branches) != 1 {
+		return nil, fmt.Errorf("Cannot rehearse jobs that run over multiple branches")
```
stevekuznetsov: Why?

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+	branch := strings.TrimPrefix(strings.TrimSuffix(source.Branches[0], "$"), "^")
```
stevekuznetsov: Oh boy

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+
+	gitrefArg := fmt.Sprintf("--git-ref=%s@%s", repo, branch)
```
stevekuznetsov: Wait, why aren't we setting the `ProwJob.Spec.Refs` ?

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+	rehearsal.Spec.Containers[0].Args = append(source.Spec.Containers[0].Args, gitrefArg)
+
+	return &rehearsal, nil
+}
+
+func loadClusterConfig() (*rest.Config, error) {
```
stevekuznetsov: We can import this, right?

cmd/rehearse/main.go
```diff
@@ -0,0 +1,222 @@
...
+func submitRehearsal(job *config.Presubmit, pr *github.PullRequest, namespace string, logger *logrus.Entry) (*pjapi.ProwJob, error) {
+	pj := pjutil.NewPresubmit(*pr, "", *job, "")
+	logger.WithFields(pjutil.ProwJobFields(&pj)).Info("Submitting a new prowjob.")
+
+	config, err := loadClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("could not load cluster config")
+	}
+
+	cs, err := pjclientset.NewForConfig(config)
```
stevekuznetsov: Should be done once at startup

cmd/rehearse/main.go
petr-muller: Good to know! Still, leaving the validation in place does not hurt imho. Maybe I'll just strengthen the error message.

cmd/rehearse/main.go
petr-muller: Because currently we rehearse by adding `--git-ref`

cmd/rehearse/main.go
petr-muller: Not sure, certainly not many, if some. Maybe some periodics. Still, it's just an error check.

cmd/rehearse/main.go
petr-muller: Because it does not fit into my current program structure where one input `JobConfig` makes one rehearsal `JobConfig`. It seem's we do not have mutlibranch jobs at all, because we force by-branch sharding of job config files anyway, so meh. But if desired, this won't be hard to fix, provided we can actually extract some meaningful branch name from the regex.

Also. any ideas about what to do with the theoretical jobs withhout the `Brancher`, meaning they run on all branches? We could always ask GitHub for branches for that repo, but I wonder if it's worth the effort to handle all these cases we do not use in practice.

cmd/rehearse/main.go
petr-muller: Cute, ain't it? The items in that list are regexes, so in some cases we might not be able to extract the "proper" name from it without asking GH for a list of all branches on the target repo and checking whether the job would run on that branch via the regex. You think it's worth the effort when we actually force our org to have a single brach with `^master$`-like regexes?

cmd/rehearse/main.go
petr-muller: Because I do not know the full impact of messing with these `ProwJob` fields. Wouldn't it break the attachment of these jobs to the `openshift/release` PRs? I don't know. If setting `ProwJob.Spec.Refs` is only used to pass the `$JOB_SPEC` to the job, then setting it is really the way to do it, thanks!

cmd/rehearse/main.go
petr-muller: Now when it lives in `ci-operator`, then yeah I guess. Probably after I extract it to some shared code location.

cmd/rehearse/main.go
petr-muller: Yeah it's in the WIP items. I was in a hurry yesterday and wanted to get this out.

cmd/rehearse/main.go
stevekuznetsov: I think the Prow library will fail to load the YAML if they have them

cmd/rehearse/main.go
stevekuznetsov: For those we should just run `master` branch. Seems fine for us to assume one branch

cmd/rehearse/main.go
stevekuznetsov: No. Agreed we can just handle the single branch cases because those are the interesting ones

cmd/rehearse/main.go
stevekuznetsov: We should be able to set the `Spec.Refs` to change the _main_ refs used for the test, which will be the ones that get used for reporting etc. We can use `extra_refs` to set other things we want to clone, like the repo under test. But not sure, that would mess up the dir that the test commands run in. Hrm.

petr-muller:
I have added a feature where a "rehearsal" ConfigMap with ci-operator configs is prepared (it's basically a rehearsal version of `ci-operator-configs` ConfigMap) for the rehearsal jobs to use. This covers the most common use case of adding new jobs (put new test to ciop config file and run the generator) and makes it work (see https://deck-ci.svc.ci.openshift.org/?job=rehearse-2367-pull-ci-openshift-ci-operator-master-build).

There is no cleanup yet, and the code is becoming a mess with all the dependencies being passed through the call tree via params, so I'll clean up everything tomorrow.

cmd/rehearse/main.go
```diff
@@ -0,0 +1,313 @@
...
+	} else {
+		logger.Fatal("Cannot start secrets agent without GitHub token")
+	}
+
+	githubClient, err := o.github.GitHubClient(secretAgent, o.dryRun)
+	if err != nil {
+		logger.WithError(err).Fatal("Failed to create GitHub client")
+	}
+
+	pr, err := githubClient.GetPullRequest(o.prowConfigOrg, o.prowConfigRepo, o.prowConfigPR)
```
droslean: Why we need this? The PR will be already `decorated` in the pod when the `pj-rehearse` job will run.

cmd/rehearse/main.go
```diff
@@ -0,0 +1,313 @@
...
+	if err != nil {
+		logger.WithError(err).WithFields(prFields).Fatal("Failed to fetch PR info from GitHub")
+	}
+
+	configAgent := &config.Agent{}
+	if err := configAgent.Start(o.configPath, o.jobConfigPath); err != nil {
```
droslean: This is totally unnecessary. Please check if you can raw load the prow config without the agent.

cmd/rehearse/main.go
droslean: What info do we need here that the GitHub client can give us that the information that we can get from $JOB_SPEC?

cmd/rehearse/main.go
petr-muller: Still, it does not hurt to check. I also realized this will be run over candidate jobs, where people can put all kinds of "invalid" jobs, so being defensive really does not hurt - if some one makes a PR with a multibranch job, we will tell him no to do it, but it's better that the rehearsal job simply refuses to rehearse it instead of blowing up on it.

cmd/rehearse/main.go
petr-muller: I vote to keep the `--git-ref` way for now to get the feature working and if needed, we can optimize the internals later.

cmd/rehearse/main.go
petr-muller: Good point. I'm using the client only to create the `PullRequest` struct that is needed by `NewPresubmit` call in `submitRehearsal`, and it is likely that `JOB_SPEC` indeed contains the same data. If it is possible to create that struct easily from `JOB_SPEC`, then we really would not need the client. I'll check it out.

petr-muller: Done!

cmd/rehearse/main.go
petr-muller: Done.

cmd/rehearse/main.go
droslean: In case the job is not running a `ci-operator` we can manipulate the `$JOB_SPEC` and pass it to the generated prowjob.

cmd/rehearse/main.go
droslean: If someone will do that, the `determine-jobs` job will fail, right? In any case, shall we make the `pj-rehearser` tool to wait for the the `determine-jobs` to be completed and then continue? or even better we can run those checks in `pj-rehearser` binary instead and get rid of the `determinize-prow-jobs` binary.

cmd/rehearse/main.go
```diff
@@ -0,0 +1,331 @@
...
+		configMapName: name,
+		neededConfigs: map[string]string{},
+	}
+}
+
+// If a job uses the `ci-operator-config` ConfigMap, save which key does it use
+// from it and replace that ConfigMap reference with a reference to the
+// temporary, rehearsal ConfigMap containing the necessary keys with content
+// matching the version from tested PR
+func (c *rehearsalCIOperatorConfigs) FixupJob(job *config.Presubmit, repo string) {
```
droslean: But all the keys will be mounted in the pod that `pj-rehearser` will run, right?

cmd/rehearse/main.go
```diff
@@ -0,0 +1,331 @@
...
+	if c.dry {
+		cmAsYAML, err := yaml.Marshal(cm)
+		if err != nil {
+			return fmt.Errorf("Failed to marshal ConfigMap to YAML: %v", err)
+		}
+		fmt.Printf("%s\n", cmAsYAML)
+		return nil
+	}
+	c.logger.Info("Creating rehearsal ConfigMap for ci-operator configs")
+	_, err := c.cmclient.Create(cm)
```
droslean: Since this will be temporary, can we somehow clear it after the jobs will finish?

petr-muller: Certainly. Currently, I do not do so because I can only delete it after all jobs that use it finish, and waiting for jobs to finish is @bbguimaraes 's turf.

cmd/rehearse/main.go
petr-muller: Not sure I understand the question. This temporary CM only contains the subset of keys that the jobs being rehearsed need, and each individual job only mounts the keys it needs. The rehearsal jobs mount the same keys as "production" jobs, just not from the original `ci-operator-config` CM, but from the rehearsal CM.

petr-muller: I plan to cover this code by unit tests that might provide some helpful examples of how the jobs are fixed up and how the resulting CM looks like.

cmd/rehearse/main.go
petr-muller: See the thread https://github.com/openshift/ci-operator-prowgen/pull/49#discussion_r243037220

We can certainly explore this, but at the moment directing the generated Prowjobs's `Refs` to point to a different repo would break the already working context reporting to the `openshift/release` PR.

petr-muller:
So it seems I have finally managed to update the vendored deps so everything builds and we are using a current version of `test-infra` here. Two downsides:

1. It needed an `[[override]]` for `Azure/go-autorest` to get rid of some build problems unrelated to our code. I haven't really dig deep into this and just found something promising on the internet and it worked.
2. The `k8s.io/test-infra/prow/config` structs used by Prowgen changed a bit in the meantime (elements common for both `Presubmit` and `Postsubmit` configs were extracted to a `JobBase` struct) which needed some adjustments in Prowgen itself. Unfortunately the changes also slightly changed the way how the job configs are marshalled in YAML, so the update will also need a big `openshift/release` PR:

```
     decorate: true
+    decoration_config:
+      skip_cloning: true
     labels:
       artifacts: images
     name: branch-ci-super-duper-release-3.11-images
-    skip_cloning: true
```

I will extract the vendor bump to a separate PR.

petr-muller:
Simple vendored deps bump PR: https://github.com/openshift/release/pull/2527

cmd/pj-rehearse/main.go
```diff
@@ -170,10 +170,10 @@ func gatherOptions() options {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 
-	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to Prow config.yaml")
-	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
+	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to *master* Prow config.yaml")
+	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to *master* Prow Prow job configs.")
 
-	fs.StringVar(&o.ciopConfigsPath, "ci-operator-configs", "", "Path to a directory containing ci-operator configs")
+	fs.StringVar(&o.candidatePath, "candidate-path", "./", "Path to a openshift/release working copy with a revision to be tested")
```
stevekuznetsov: Not a huge fan of default values for these -- let's be explicit wherever possible

pkg/diffs/diffs.go
```diff
@@ -2,21 +2,27 @@ package diffs
 
 import (
 	"fmt"
+	"path/filepath"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	prowconfig "k8s.io/test-infra/prow/config"
 )
 
+const configInRepoPath = "cluster/ci/config/prow/config.yaml"
```
stevekuznetsov, line 11: This as well -- would love to see it come in with flags

stevekuznetsov:
@petr-muller can you rebase this on the deps PR?

cmd/pj-rehearse/main.go
petr-muller: I am an explicit fan too. I wanted to respect how Nikos originally did it. I can remove the default but the existing presubmit will need to be adjusted.

pkg/diffs/diffs.go
petr-muller: With these I am not that sure. These are just paths into `openshift/release` directory, to which the tool is already tied to anyway, so I do not see a problem with hardcoding these because they won't change often.

We already need three such paths and we will need more once we start handling templates, which would be four paths to `openshift/release` which we would pass as parameters but they are basically constants.

petr-muller:
> @petr-muller can you rebase this on the deps PR?

Done.

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	"encoding/json"
 	"flag"
 	"fmt"
-	"log"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	// TODO: Solve this properly
```
stevekuznetsov: What is this and have we solved it?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/ci-operator-prowgen/pkg/diffs"
 )
 
+func getPrNumber(jobSpec *pjapi.ProwJobSpec) int {
```
stevekuznetsov: Why is this useful as a function?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	return jobSpec.Refs.Pulls[0].Number
```
stevekuznetsov: Not guaranteed to have `Pulls` filled in, generally -- do we know this is only called for presubmits or batches?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+}
+
+func getJobSpec() (*pjapi.ProwJobSpec, error) {
```
stevekuznetsov: I swear we had this logic in `test-infra` `pod-utilities` or similar

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	if len(specEnv) == 0 {
+		return nil, fmt.Errorf("JOB_SPEC not set or set to an empty string")
+	}
+	spec := pjapi.ProwJobSpec{}
+	if err := json.Unmarshal(specEnv, &spec); err != nil {
+		return nil, err
+	}
+
+	if len(spec.Refs.Pulls) > 1 {
+		return nil, fmt.Errorf("Cannot rehearse in the context of a batch job")
```
stevekuznetsov: Prefer to check the job type directly

stevekuznetsov: Also, why?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	configDir string
+
+	logger logrus.FieldLogger
+	dry    bool
+
+	configMapName string
+	neededConfigs map[string]string
+}
+
+func newRehearsalCIOperatorConfigs(cmclient corev1.ConfigMapInterface, prNumber int, configDir string, logger logrus.FieldLogger, dry bool) *rehearsalCIOperatorConfigs {
```
stevekuznetsov: If this is an unexported constructor, the callers could just set the fields themselves

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	if c.dry {
+		cmAsYAML, err := yaml.Marshal(cm)
+		if err != nil {
+			return fmt.Errorf("Failed to marshal ConfigMap to YAML: %v", err)
+		}
+		fmt.Printf("%s\n", cmAsYAML)
+		return nil
+	}
+	c.logger.Info("Creating rehearsal ConfigMap for ci-operator configs")
+	_, err := c.cmclient.Create(cm)
```
stevekuznetsov: Do we have any cleanup yet?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	return err
+}
+
+func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber int) (*prowconfig.Presubmit, error) {
+	var rehearsal prowconfig.Presubmit
+	deepcopy.Copy(&rehearsal, source)
```
stevekuznetsov: `ProwJob` has a method `DeepCopy()` -- use that

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	rehearsal.Name = fmt.Sprintf("rehearse-%d-%s", prNumber, source.Name)
+	rehearsal.Context = fmt.Sprintf("ci/rehearse/%s/%s", repo, strings.TrimPrefix(source.Context, "ci/prow/"))
+
+	if len(source.Spec.Containers) != 1 {
+		return nil, fmt.Errorf("Cannot rehearse jobs with more than 1 container in Spec")
+	}
+	container := source.Spec.Containers[0]
+
+	if len(container.Command) != 1 || container.Command[0] != "ci-operator" {
+		return nil, fmt.Errorf("Cannot rehearse jobs that have Command different from simple 'ci-operator'")
```
stevekuznetsov: Will this error stop other jobs from being created?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+func submitRehearsal(job *prowconfig.Presubmit, jobSpec *pjapi.ProwJobSpec, logger logrus.FieldLogger, pjclient pj.ProwJobInterface, dry bool) (*pjapi.ProwJob, error) {
+	labels := make(map[string]string)
+	for k, v := range job.Labels {
+		labels[k] = v
+	}
+
+	pj := pjutil.NewProwJob(pjutil.PresubmitSpec(*job, *(jobSpec.Refs)), labels)
+	logger.WithFields(pjutil.ProwJobFields(&pj)).Info("Submitting a new prowjob.")
+
+	if dry {
```
stevekuznetsov: Can we do this in a wrapper of the client?

cmd/pj-rehearse/main.go
```diff
@@ -1,23 +1,278 @@
...
+	}
+
+	if len(rehearsals) > 0 {
+		if err := rehearsalConfigs.Create(); err != nil {
+			return fmt.Errorf("failed to prepare rehearsal ci-operator config ConfigMap: %v", err)
+		}
+		for _, job := range rehearsals {
+			created, err := submitRehearsal(job, jobSpec, logger, pjclient, dry)
+			if err != nil {
+				logger.WithError(err).Warn("Failed to execute a rehearsal presubmit presubmit")
```
stevekuznetsov: stutter

cmd/pj-rehearse/main.go
```diff
@@ -32,19 +287,54 @@ func main() {
...
-		log.Fatal(err)
+		logrus.WithError(err).Fatal("invalid options")
+	}
+
+	jobSpec, err := getJobSpec()
+	if err != nil {
+		logrus.WithError(err).Fatal("could not read JOB_SPEC")
+	}
+
+	prFields := logrus.Fields{"org": jobSpec.Refs.Org, "repo": jobSpec.Refs.Repo, "PR": getPrNumber(jobSpec)}
```
stevekuznetsov: There's a helper upstream that gives you fields for spec

pkg/rehearse/configmaps_test.go
```diff
@@ -0,0 +1,199 @@
...
+	if ok {
+		return string(content), nil
+	}
+
+	return "", fmt.Errorf("Failed to read a file %s", path)
+}
+
+func TestCreate(t *testing.T) {
+	testLogger := logrus.New()
+	testLogger.SetOutput(ioutil.Discard)
```
stevekuznetsov: Seeing logs from tests is often useful

Gopkg.toml
```diff
@@ -1,10 +1,24 @@
...
+	"k8s.io/test-infra/prow/apis/prowjobs/v1",
+	"k8s.io/test-infra/prow/client/clientset/versioned",
+	"k8s.io/test-infra/prow/client/clientset/versioned/typed/prowjobs/v1",
+
 	"k8s.io/test-infra/prow/kube",
+	"k8s.io/test-infra/prow/config",
+	"k8s.io/test-infra/prow/pjutil",
+
+	"k8s.io/client-go/tools/clientcmd",
+	"k8s.io/client-go/kubernetes/typed/core/v1",
```
stevekuznetsov, line 18: Why didn't this fall out in the rebase?

pkg/rehearse/jobs.go
```diff
@@ -0,0 +1,106 @@
+package rehearse
```
stevekuznetsov, line 1: This could use some unit testing

stevekuznetsov (commented):
/lint

cmd/pj-rehearse/main.go
petr-muller: Importing a lib for deepcopy felt like a temporary workaround. It's not because prowconfig.Presubmit & friends do not have DeepCopy themselves.

cmd/pj-rehearse/main.go
petr-muller: It's more readable at callsite than the ugly expression it wraps.

cmd/pj-rehearse/main.go
petr-muller: Yes, this is supposed to be always called within the rehearsal-spawning job, which is always a presubmit on `openshift/release` PR.

petr-muller: "Handle batches" is basically still a TODO, current implementation does not deal with them properly and I would like to implement it as an increment.

cmd/pj-rehearse/main.go
petr-muller: Thanks I'll have a look!

cmd/pj-rehearse/main.go
petr-muller: Hmm... I'll need to have a better look at what's in the JobSpec, haven't noticed there's a job type except the pre/postsubmit etc.

"Handle batches" is basically still a TODO, current implementation does not deal with them properly and I would like to implement it as an increment.

cmd/pj-rehearse/main.go
petr-muller: Right, thanks. Evolution artifact :)

cmd/pj-rehearse/main.go
petr-muller: Not yet, current version just spawns the jobs and exits, it cannot clean up these CMs at that moment because some ProwJob-created Pods might not even be running yet. And waiting on jobs to finish is @bbguimaraes' turf.

We'll eventually want to add some cleanup even to the startup, but that's an increment.

cmd/pj-rehearse/main.go
petr-muller: It's not a `ProwJob`, it's a `Presubmit` from https://github.com/kubernetes/test-infra/blob/master/prow/config/config.go It does not have it's own `DeepCopy()` :(

The lib does not seem to be that big hit to be vendored, compared to all that :hankey: that we need from kube brings :)

cmd/pj-rehearse/main.go
petr-muller: Nope, it's just logged as a warning at callsite where all to-be-rehearsed jobs are processed in a loop.

cmd/pj-rehearse/main.go
petr-muller: I guess we can, for both clients we currently use for modifying calls.

cmd/pj-rehearse/main.go
petr-muller: Thanks, I'll have a look!

Gopkg.toml
petr-muller: Not sure I understand the question. We use this.

pkg/rehearse/jobs.go
petr-muller: Yeah, I'll create some more tomorrow.

pkg/rehearse/configmaps.go
```diff
@@ -0,0 +1,132 @@
...
+// the tested PR. All ci-operator-based jobs use the `ci-operator-configs`
+// ConfigMap that contains ci-operator configuration files. Rehearsed jobs
+// need to have the PR-version of these files available. The following code
+// takes care of creating a short-lived, rehearsal ConfigMap. The keys needed
+// to be present are extracted from the rehearsal jobs and the rehearsal jobs
+// are modified to use this ConfigMap instead of the "production" one.
+
+var ciOperatorConfigsCMName = "ci-operator-configs"
+
+type CIOperatorConfigs interface {
```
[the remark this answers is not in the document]
petr-muller: Bloody pedantic bot :/

cmd/pj-rehearse/main.go
stevekuznetsov: `prowjob.spec.type` is the field you want

cmd/pj-rehearse/main.go
stevekuznetsov: `pjutil.ProwJobFields`

Gopkg.toml
stevekuznetsov: I thought you put in the deps first, though?

cmd/pj-rehearse/main.go
stevekuznetsov: :man_facepalming: I see

cmd/pj-rehearse/main.go
petr-muller: Indeed, `pod-utils/downwardapi.ResolveSpecFromEnv`.

cmd/pj-rehearse/main.go
petr-muller: Huh, I totally misunderstood that `batch` is a separate job type on it's own. I lived under impression that a batch job is just a presubmit with multiple PRs.

cmd/pj-rehearse/main.go
petr-muller: Unfortunately that takes `ProwJob` and not `ProwJobSpec` :( We **could** make the former from the latter and then use that, but IMO that's not worth it just to extract few values.

Gopkg.toml
petr-muller: The first (already merged) PR only updated what we already vendored for Prowgen itself, mostly `test-infra` stuff and its deps. Looks like these were previously only pulled transitively and were added to `Gopkg.toml` because rehearse use these directly.

cmd/pj-rehearse/main.go
petr-muller: It's now gone, part of https://github.com/openshift/ci-operator-prowgen/pull/49/commits/edc2713ecb3953507120117d44168f72e8f7a8fe

cmd/pj-rehearse/main.go
petr-muller: In https://github.com/openshift/ci-operator-prowgen/pull/49/commits/96882cdc8b030d7435dba65aa1555125b08b668b I now explicitly check for the job being a presubmit and bail out if otherwise. Also, after https://github.com/openshift/ci-operator-prowgen/pull/49/commits/edc2713ecb3953507120117d44168f72e8f7a8fe the prNumber is now only extracted once and just the fields from the spec that we actually need are passed around, instead of full `jobSpec`

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/edc2713ecb3953507120117d44168f72e8f7a8fe

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/96882cdc8b030d7435dba65aa1555125b08b668b

cmd/pj-rehearse/main.go
petr-muller: Adressed already.

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/08b1aa53db5d54f3cbdaf2a11f0024a2b2b6c697

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/d8f71b4ffd2d83717659ce733846c3519d521e86

pkg/rehearse/configmaps_test.go
petr-muller: https://github.com/openshift/ci-operator-prowgen/pull/49/commits/d8f71b4ffd2d83717659ce733846c3519d521e86

petr-muller (commented):
/lint

pkg/rehearse/configmaps.go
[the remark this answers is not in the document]
petr-muller: https://github.com/openshift/ci-operator-prowgen/pull/49/commits/d8f71b4ffd2d83717659ce733846c3519d521e86

pkg/rehearse/jobs.go
```diff
@@ -0,0 +1,106 @@
...
+			return nil, fmt.Errorf("Failed to marshal job to YAML: %v", err)
+		}
+		fmt.Printf("%s\n", jobAsYAML)
+		return &pj, nil
+	}
+
+	return pjclient.Create(&pj)
+}
+
+func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, jobSpec *pjapi.ProwJobSpec, logger logrus.FieldLogger, rehearsalConfigs CIOperatorConfigs, pjclient pj.ProwJobInterface, dry bool) error {
```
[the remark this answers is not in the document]
petr-muller: https://github.com/openshift/ci-operator-prowgen/pull/49/commits/d8f71b4ffd2d83717659ce733846c3519d521e86

petr-muller:
At this moment, I think all review comments with exception of additional tests for `jobs.go` are addressed, I have tried to respond to all comments and add sausage-like small commits to make followup reviews easier.

I'll now work on the additional tests.

droslean:
It makes more sense to put `diffs` in `rehearse` package as well.

cmd/pj-rehearse/main.go
stevekuznetsov: Could refactor `ProwJobFields` to call `ProwJobSpecFields` and use that, idk

stevekuznetsov: Very useful to have standard fields in logging for machine-driven filtering etc

stevekuznetsov: Using identical field names here without refactor SGTM for now

cmd/pj-rehearse/main.go
```diff
@@ -109,6 +105,14 @@ func main() {
 
 	prFields := logrus.Fields{"org": jobSpec.Refs.Org, "repo": jobSpec.Refs.Repo, "PR": getPrNumber(jobSpec)}
 	logger := logrus.WithFields(prFields)
+
+	if jobSpec.Type != "presubmit" {
```
stevekuznetsov: That should be a typed constant to import

cmd/pj-rehearse/main.go
```diff
@@ -113,6 +95,9 @@ func main() {
 		os.Exit(0)
 	}
 
+	prNumber := jobSpec.Refs.Pulls[0].Number
+	logger = logrus.WithField("PR", prNumber)
```
stevekuznetsov: This isn't a field we use upstream

stevekuznetsov (commented):
Nice!

pkg/rehearse/configmaps.go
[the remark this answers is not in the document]
droslean: Why use an interface here?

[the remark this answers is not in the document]
petr-muller: Less coupling. The other parts do not need to care about internals, which also makes testing simpler.

cmd/rehearse/main.go
droslean: Btw, we can do that without creating a configMap for it. Since that the `pj-rehearse` will actually create the prowjob, it can just put the config as value instead of mounting the config from a configMap that we actually create it in the first place.

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/c8b379bfea9953128bf5db1957e6822b5091829e

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/c8b379bfea9953128bf5db1957e6822b5091829e

cmd/pj-rehearse/main.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/49/commits/c8b379bfea9953128bf5db1957e6822b5091829e by using names for the fields imported from `test-infra`.

pkg/rehearse/jobs.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil
+}
+
+// Rehearsed Prow jobs may depend on ConfigMaps with content also modified by
+// the tested PR. All ci-operator-based jobs use the `ci-operator-configs`
+// ConfigMap that contains ci-operator configuration files. Rehearsed jobs
+// need to have the PR-version of these files available. The following code
+// takes care of creating a short-lived, rehearsal ConfigMap. The keys needed
+// to be present are extracted from the rehearsal jobs and the rehearsal jobs
+// are modified to use this ConfigMap instead of the "production" one.
```
droslean: nit: The comment above seems outdated too.

PR #50, opened 2019-01-04
WIP Initial PR for checking which preSubmits to be executed.

droslean:
With this PR, I add the logic of how to compare the files of the decorated repo and the master. We consider the master files mounted from the configMaps in the Prowjob. 
The `diffs` package will provide a `map[string][]Presubmit` with the Presubmits that are required to be executed. Any changes that are not related to the `.spec` of the job, will be ignored.  

Also, the `main.go` will be implemented in #49

WIP because tests are missing.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+	"net/url"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/diffs"
+)
+
+type options struct {
+	ciOpConfigPath  string
+	jobConfigPath   string
+	decorationPath  string
+	githubEndpoint  string
+	githubTokenFile string
+}
+
+func gatherOptions() options {
+	o := options{}
+	flag.StringVar(&o.githubEndpoint, "github-endpoint", "https://api.github.com", "GitHub's API endpoint.")
+	flag.StringVar(&o.githubTokenFile, "github-token-file", "", "Path to file containing GitHub OAuth token.")
+	flag.StringVar(&o.ciOpConfigPath, "ci-op-config-path", "", "Path to ci-operator's configuration files..")
+	flag.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
+	flag.StringVar(&o.decorationPath, "decoration-path", "./", "Path where the repository has been cloned.")
+	flag.Parse()
+	return o
+}
+
+func validateOptions(o options) error {
+	if len(o.githubTokenFile) == 0 {
+		return fmt.Errorf("empty --github-token-file")
+	}
+
+	if len(o.githubEndpoint) == 0 {
+		return fmt.Errorf("empty --github-endpoint")
+	} else if _, err := url.Parse(o.githubEndpoint); err != nil {
+		return fmt.Errorf("bad --github-endpoint provided: %v", err)
+	}
+
+	if len(o.ciOpConfigPath) == 0 {
+		return fmt.Errorf("empty --ci-op-config-path")
+	}
+
+	if len(o.jobConfigPath) == 0 {
+		return fmt.Errorf("empty --job-config-path")
+	}
+
+	return nil
+}
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(o)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	diffs := diffs.NewDiffs(o.jobConfigPath, o.ciOpConfigPath, o.decorationPath)
+	//diffs.GetPresubmitsToExecute()
+	preSubmitsToExecute := diffs.GetPresubmitsToExecute()
+
+	// // Just print the map with the presubmits to be executed.
+	// // TODO: execute them
+	for k, v := range preSubmitsToExecute {
+		log.Printf("############### %s ###############:", k)
+		for _, p := range v {
+			log.Printf("%s", p.Name)
+		}
+	}
+}
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -1,0 +1,140 @@
+package diffs
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"k8s.io/apimachinery/pkg/api/equality"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+// PreSubmits holds the logic of prow's presubmits yaml format.
+type PreSubmits struct {
+	PreSubmits map[string][]prowconfig.Presubmit `yaml:"presubmits,omitempty"`
+}
+
+// Diffs holds the logic about the differences between the files.
+type Diffs struct {
+	JobConfigPath  string
+	CiOpConfigPath string
+	DecorationPath string
+}
+
+// NewDiffs creates a new Diffs{}
+func NewDiffs(jobConfigPath, ciOpConfigPath, decorationPath string) *Diffs {
+	return &Diffs{
+		JobConfigPath:  jobConfigPath,
+		CiOpConfigPath: ciOpConfigPath,
+		DecorationPath: decorationPath,
+	}
+
+}
+
+func (d *Diffs) findChangedFiles() ([]string, error) {
+	var changedFiles []string
+	// Walk through the cloned repo
+	if err := filepath.Walk(path.Join("ci-operator/jobs/", d.DecorationPath), func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.Printf("prevent panic by handling failure accessing a path %q: %v\n", path, err)
+			return err
+		}
+
+		if !info.IsDir() && filepath.Ext(path) == ".yaml" {
+			masterFileContents, err := ioutil.ReadFile(filepath.Join(d.JobConfigPath, filepath.Base(path)))
+			if err != nil {
+				log.Printf("Can't read master's file: %v", err)
+			}
+			prFileContents, err := ioutil.ReadFile(path)
+			if err != nil {
+				log.Printf("Error reading file %s: %v", path, err)
+			}
+			if bytes.Compare(masterFileContents, prFileContents) != 0 {
+				changedFiles = append(changedFiles, path)
+			}
+		}
+		return nil
+	}); err != nil {
+		return nil, err
+	}
+	return changedFiles, nil
+}
+
+func (d *Diffs) getPresubmitChanges() []string {
+	var preSubmitChangedFiles []string
+
+	changedFiles, err := d.findChangedFiles()
+	if err != nil {
+		log.Fatal("Couldn't get changed files.")
+	}
+
+	for _, changedFile := range changedFiles {
+		if strings.Contains(filepath.Base(changedFile), "presubmits.yaml") {
+			preSubmitChangedFiles = append(preSubmitChangedFiles, changedFile)
+		}
+	}
+	return preSubmitChangedFiles
+}
+
+// We return a map of maps here to make the comparing more efficient.
+func getPresubmits(file string) (map[string]map[string]prowconfig.Presubmit, error) {
+	var preSubmits PreSubmits
+
+	jobsByRepo := make(map[string]map[string]prowconfig.Presubmit)
+
+	data, err := ioutil.ReadFile(file)
+	if err != nil {
+		return jobsByRepo, fmt.Errorf("failed to read presubmit file (%v)", err)
+	}
+
+	if err := yaml.Unmarshal(data, &preSubmits); err != nil {
+		return jobsByRepo, fmt.Errorf("failed to unmarshal presubmit file (%v)", err)
+	}
+
+	for repo, preSubmitList := range preSubmits.PreSubmits {
+		pm := make(map[string]prowconfig.Presubmit)
+		for _, p := range preSubmitList {
+			pm[p.Name] = p
+			jobsByRepo[repo] = pm
+		}
+	}
+	return jobsByRepo, nil
+}
+
+// GetPresubmitsToExecute returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetPresubmitsToExecute() map[string][]prowconfig.Presubmit {
+	var preSubmitsToExecute []prowconfig.Presubmit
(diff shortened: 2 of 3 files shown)


pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+		JobConfigPath:  jobConfigPath,
+		CiOpConfigPath: ciOpConfigPath,
+		DecorationPath: decorationPath,
+	}
+
+}
+
+func (d *Diffs) findChangedFiles() error {
+	// Walk through the cloned repo
+	if err := filepath.Walk(fmt.Sprintf("%s/ci-operator/jobs/", d.DecorationPath), func(path string, info os.FileInfo, err error) error {
```
stevekuznetsov: We want `path.Join` instead of `fmt.Sprintf`

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+		if err != nil {
+			log.Printf("prevent panic by handling failure accessing a path %q: %v\n", path, err)
+			return err
+		}
+
+		if !info.IsDir() && filepath.Ext(path) == ".yaml" {
+			masterFileContents, err := ioutil.ReadFile(fmt.Sprintf("%s/%s", d.JobConfigPath, filepath.Base(path)))
```
stevekuznetsov: Instead of `filepath.Base` don't we want `filepath.Rel` from the `path` to the thing we're walking through?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+func NewDiffs(jobConfigPath, ciOpConfigPath, decorationPath string) *Diffs {
+	return &Diffs{
+		JobConfigPath:  jobConfigPath,
+		CiOpConfigPath: ciOpConfigPath,
+		DecorationPath: decorationPath,
+	}
+
+}
+
+func (d *Diffs) findChangedFiles() error {
```
stevekuznetsov: Instead of having a side effect of populating a member of this struct, why not make it into a pure function and return the changed files? Easier to test

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+	return nil
+}
+
+func (d *Diffs) getPresubmitChanges() []string {
+	var preSubmitChangedFiles []string
+	if err := d.findChangedFiles(); err != nil {
+		log.Fatal("Couldn't populate changed files.")
+	}
+	for _, changedFile := range d.ChangedFiles {
+		if strings.Contains(filepath.Base(changedFile), "presubmits.yaml") {
```
stevekuznetsov, line 78: Shouldn't we load the file and see what is in it? Would be more robust

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+			preSubmitChangedFiles = append(preSubmitChangedFiles, changedFile)
+		}
+	}
+	return preSubmitChangedFiles
+}
+
+// We return a map of maps here to make the comparing more efficient.
```
stevekuznetsov, line 85: How?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+func getPresubmits(file string) (map[string]map[string]prowconfig.Presubmit, error) {
+	var preSubmits PreSubmits
+
+	preSubmitsMapOfMaps := make(map[string]map[string]prowconfig.Presubmit)
```
stevekuznetsov: maybe `jobsByRepo`? The mapping is job type --> job repo --> job here

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+	for repo, preSubmitList := range preSubmits.PreSubmits {
+		for _, p := range preSubmitList {
+			pm[p.Name] = p
+			preSubmitsMapOfMaps[repo] = pm
+		}
+	}
+	return preSubmitsMapOfMaps, nil
+}
+
+// GetPresubmitsToExecute returns a map of strings with a list of Presubmits to be executed.
```
stevekuznetsov: Instead of `"map of strings with a list"` maybe `"returns a mapping of repo to presubmits to execute"`?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+func (d *Diffs) GetPresubmitsToExecute() map[string][]prowconfig.Presubmit {
+	var preSubmitsToExecute []prowconfig.Presubmit
+	preSubmitsToExecuteMap := make(map[string][]prowconfig.Presubmit)
+
+	if preSubmitChangedFiles := d.getPresubmitChanges(); len(preSubmitChangedFiles) > 0 {
```
stevekuznetsov: Not sure you need the `len` check, the following `for` will do nothing if it's empty

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+		for _, preSubmitChangedFile := range preSubmitChangedFiles {
+			prPreSubmits, _ := getPresubmits(preSubmitChangedFile)
```
stevekuznetsov: Why ignore errors here?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+				preSubmitsToExecute = []prowconfig.Presubmit{}
+				for jobName, job := range jobs {
+					if !equality.Semantic.DeepEqual(masterPreSubmits[repo][jobName].Spec, job.Spec) {
+						preSubmitsToExecute = append(preSubmitsToExecute, job)
+					}
+				}
+
+				// The key can already exist if a different branch has already been checked in.
+				if _, ok := preSubmitsToExecuteMap[repo]; ok {
+					for _, job := range preSubmitsToExecute {
```
stevekuznetsov: Instead of appending one by one you can 

```go
preSubmitsToExecuteMap[repo] = append(preSubmitsToExecuteMap[repo], preSubmitsToExecute...)
```

stevekuznetsov (commented):
I'm worried this assumes that a job doesn't move between files. While this may normally be true, it may be better to load the full config at once from both locations and run the diff on the merged config for each version.

pkg/diffs/diffs.go
petr-muller: It makes referencing jobs *by name* easier, I guess. test-infra config structs have plain lists, it makes searching for a jobs by name hard.

cmd/pj-rehearse/main.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/openshift/ci-operator-prowgen/pkg/diffs"
+)
+
+type options struct {
+	ciOpConfigPath  string
+	jobConfigPath   string
+	decorationPath  string
+	githubEndpoint  string
+	githubTokenFile string
```
petr-muller, line 17: `ciOpConfigPath`, `githubEndpoint` and `githubTokenFile` are not used anywhere. I think we should add code as-needed, so it's clear in what context is it used.

I am not also not that happy with the `decorationPath` name, as it does not really describe what it is. I think we should accept symmetrical options for master and under-test configs. Either accept paths to full `openshift/release` repo clones (one for master, one for PR) and compute the `jobConfig` paths by knowing the layout of `openshift/release`, or accept two sets of options (again, one for master, one for pr) pointing to needed directories directly.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+type PreSubmits struct {
+	PreSubmits map[string][]prowconfig.Presubmit `yaml:"presubmits,omitempty"`
+}
+
+// Diffs holds the logic about the differences between the files.
+type Diffs struct {
+	JobConfigPath  string
+	CiOpConfigPath string
+	DecorationPath string
+	ChangedFiles   []string
```
petr-muller: `ChangedFiles` seems to be used only as a faux (indirect) return value of `findChangedFiles`, maybe that method should just straight return the list and this member would not be needed.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+		for _, p := range preSubmitList {
+			pm[p.Name] = p
+			preSubmitsMapOfMaps[repo] = pm
+		}
+	}
+	return preSubmitsMapOfMaps, nil
+}
+
+// GetPresubmitsToExecute returns a map of strings with a list of Presubmits to be executed.
+func (d *Diffs) GetPresubmitsToExecute() map[string][]prowconfig.Presubmit {
```
petr-muller, line 111: I think do not assume what the Presubmits are for. They are Presubmits affected by a proposed change, what they are used for is not this code's business. The `...ToExecute` suffix increases the code coupling unnecessarily.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Diffs) getPresubmitChanges() []string {
```
petr-muller, line 69: The name is confusing here, it returns a list of *file paths* that differ, not any changes.

pkg/diffs/diffs.go
petr-muller: Agreed, looked fishy to me too. It also needs `filepath.Join` instead of that `Sprintf` here.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+
+// GetPresubmitsToExecute returns a map of strings with a list of Presubmits to be executed.
+func (d *Diffs) GetPresubmitsToExecute() map[string][]prowconfig.Presubmit {
+	var preSubmitsToExecute []prowconfig.Presubmit
+	preSubmitsToExecuteMap := make(map[string][]prowconfig.Presubmit)
+
+	if preSubmitChangedFiles := d.getPresubmitChanges(); len(preSubmitChangedFiles) > 0 {
+		for _, preSubmitChangedFile := range preSubmitChangedFiles {
+			prPreSubmits, _ := getPresubmits(preSubmitChangedFile)
+			masterPreSubmits, _ := getPresubmits(fmt.Sprintf("%s/%s", d.JobConfigPath, filepath.Base(preSubmitChangedFile)))
```
petr-muller: I think this will suffer from the same things like here https://github.com/openshift/ci-operator-prowgen/pull/50/files#r245326988

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+
+			for repo, jobs := range prPreSubmits {
+				preSubmitsToExecute = []prowconfig.Presubmit{}
+				for jobName, job := range jobs {
+					if !equality.Semantic.DeepEqual(masterPreSubmits[repo][jobName].Spec, job.Spec) {
```
petr-muller: I think this will crash on newly added jobs. Newly added job's `jobName` won't have an appropriate master presubmit item, and we will attempt to access such item's `Spec`.

Also, are we really sure we do not want to rehearse jobs changed in non-Spec fields? Are all these fields that harmless?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+						preSubmitsToExecute = append(preSubmitsToExecute, job)
+					}
+				}
+
+				// The key can already exist if a different branch has already been checked in.
```
petr-muller: The comment mentioning a `branch` misled me for a while from what this is really for. This is needed when two different files contain presubmits for a same repo, which can really happen, but the comment should be improved to better describe it.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+	}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (d *Diffs) getPresubmitChanges() []string {
+	var preSubmitChangedFiles []string
+	if err := d.findChangedFiles(); err != nil {
+		log.Fatal("Couldn't populate changed files.")
```
petr-muller: This should not `Fatal`, but pass the error out like everything else.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	if err := yaml.Unmarshal(data, &preSubmits); err != nil {
+		return preSubmitsMapOfMaps, fmt.Errorf("failed to unmarshal presubmit file (%v)", err)
+	}
+
+	for repo, preSubmitList := range preSubmits.PreSubmits {
+		for _, p := range preSubmitList {
+			pm[p.Name] = p
+			preSubmitsMapOfMaps[repo] = pm
```
petr-muller: This is weird. This way the `pm` map just gradually grows and the same map is continuously (and needlessly) assigned to all the keys in the `preSubmitsMapOfMaps` map.

I think what needs to happen is that new `pm` needs to created for each repo, keys should be created in it and then it should be assigned to the `preSubmitMapOfMaps[repo]` once at the end, something like:

```
for repo, list := range presubmits
  pm = make(map...)
  for _, p := range list {
    pm[p.Name] = p
  preSubmitMapOfMaps[repo] = pm
}
```

petr-muller (commented):
I have a bunch of comments to the code, but my most important comment, if addressed, would render most of the code comments unneeded. I believe most of the fiddling and comparing code is not even needed. I think the detection should fully utilize the `config.Load` method from `test-infra/prow/config` (just like my sibling PR https://github.com/openshift/ci-operator-prowgen/pull/49/files#diff-3858c6d6560cc81c2d77e6c29a9632acR302 does) and simply load the full config for both master and PR versions. That way you do not need to deal at all with the raw files, and can do all comparisons needed simply by walking the `JobConfig` struct (or more precisely, it's `Presubmits` member)

pkg/diffs/diffs.go
droslean: I need to compare like `equality.Semantic.DeepEqual` but also I want to know where are the differences. There are other libraries that are trying to do that with ugly approaches but instead I just add a new key with the job name and I can easily get the job I want to compare by the job name.

petr-muller:
Plus, I'd like to point out that the agreement (also written down in the [DPTP-291](https://jira.coreos.com/browse/DPTP-291) JIRA) was that the initial version would detect newly added jobs, not changed jobs. While I understand that we will eventually want to rehearse even changed jobs, if we really want to collaborate efficiently we should stick to developing what was agreed.

pkg/diffs/diffs.go
droslean: We will have to be more specific in which fields we need to execute the job. Here I only check the `.spec`.

pkg/diffs/diffs.go
droslean: If that fails, there is no reason to continue.

PR #51, opened 2019-01-04
Watch jobs during rehearsal and report results

bbguimaraes:


cmd/rehearse/main.go
```diff
@@ -226,25 +227,62 @@ func execute(jobs config.JobConfig, jobSpec *pjapi.ProwJobSpec, logger logrus.Fi
...
-		}
-	} else {
+	if len(rehearsals) == 0 {
 		logger.Warn("No job rehearsals")
+		return nil
+	}
+	if err := rehearsalConfigs.Create(); err != nil {
+		return fmt.Errorf("failed to prepare rehearsal ci-operator config ConfigMap: %v", err)
+	}
+	watcher, err := pjclient.Watch(metav1.ListOptions{})
```
bbguimaraes: This seems to do an implicit `List`, which is completely unnecessary. I will see if it is possible to have this be some kind of API call barrier so we only get events that happen after the watch starts.

We should also add labels/annotations to the generated jobs to remove the noise from the watch.

petr-muller: Yeah, I was aware we will want *some* labels but haven't thought about how they should look like.

Would labelling all entities (not just prowjobs but also rehearsal CMs and the like) with something like `rehearsal: $PRNUMBER` make sense? Would we need some "namespace" for the label (like the `prow.k8s.io` ones have?)?

cmd/rehearse/main.go
```diff
@@ -226,25 +227,62 @@ func execute(jobs config.JobConfig, jobSpec *pjapi.ProwJobSpec, logger logrus.Fi
...
+		}
+		logger.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
+		pjs[created.Name] = struct{}{}
+	}
+	success, err := waitForJobs(pjs, watcher.ResultChan(), logger)
+	if err != nil {
+		return fmt.Errorf("error while waiting for jobs: %s", err)
+	}
+	if !success {
+		return fmt.Errorf("not all ProwJobs were successful")
```
petr-muller: I was briefly thinking about this, and I think it's worth the additional effort to think more about the failure modes here. We will probably need to easily distinguish between the "we submitted all rehearsals and some failed" (which is not actually a failure, but the rehearsal tool working as intended) and the cases where something actually failed within what the rehearsal attempts to do.

cmd/rehearse/main.go
bbguimaraes: That is what I had in mind, it would be sufficient for this purpose.

As for the namespace, we use `ci.openshift.org` in several places already (ci-operator, secret mirroring, etc.).

cmd/rehearse/main.go
bbguimaraes: Yes, that is merely a placeholder for now while I focus on the watch. Proper error handling will come later.

pkg/rehearse/configmaps.go
```diff
@@ -135,8 +136,11 @@ func (c *ciOperatorConfigs) FixupJob(job *prowconfig.Presubmit, repo string) {
 // rehearsal jobs.
 func (c *ciOperatorConfigs) Create() error {
 	cm := &v1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{Name: c.configMapName},
-		Data:       map[string]string{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:   c.configMapName,
+			Labels: map[string]string{rehearseLabel: strconv.Itoa(c.prNumber)},
```
stevekuznetsov: Don't we want this to be local to the specific batch of jobs triggered by this run?

pkg/rehearse/jobs.go
```diff
@@ -123,21 +125,82 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
 		}
-		for _, job := range rehearsals {
-			created, err := submitRehearsal(job, refs, logger, pjclient)
-			if err != nil {
-				logger.WithError(err).Warn("Failed to execute a rehearsal presubmit")
-			} else {
-				logger.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
-			}
+		logger.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
+		pjs[created.Name] = struct{}{}
```
stevekuznetsov: Use kubernetes sets.String

pkg/rehearse/jobs.go
```diff
@@ -123,21 +125,82 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+	}
+	selector := fmt.Sprintf("%s=%d", rehearseLabel, prNumber)
```
stevekuznetsov: I'd rather see this done by building the selector and using it's toString method

pkg/rehearse/jobs.go
```diff
@@ -123,21 +125,82 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+	var w watch.Interface
```
stevekuznetsov: Why do we need to have this exist outside the scope of `waitForJobs`?

pkg/rehearse/jobs.go
```diff
@@ -123,21 +125,82 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+		for {
+			event, ok := <-w
+			if !ok {
+				break
+			}
+			pj, ok := event.Object.(*pjapi.ProwJob)
+			if !ok {
+				return false, fmt.Errorf("received a %T from watch", event.Object)
+			}
+			logger.WithFields(logrus.Fields{"name": pj.Name, "state": pj.Status.State}).Info("processing ProwJob")
```
stevekuznetsov: `pjutil.ProwJobFields` would be nice here, then add in the state

pkg/rehearse/configmaps.go
bbguimaraes: The only purpose at the moment is to reduce watch traffic — jobs that match but were not created by this execution (a much smaller set) are ignored by `waitForJobs`.

The PR number was a quick way of achieving that. I could not think of a unique identifier that was (1) more specific and (2) this simple.

pkg/rehearse/jobs.go
bbguimaraes: Nice.

pkg/rehearse/jobs.go
bbguimaraes: To keep the original interface (`func() (<-chan watch.Event, error)`) before retries made things more complicated. It made testing easier.

I think a proper client interface can make this cleaner.

pkg/rehearse/configmaps.go
stevekuznetsov: Hash of the job spec would work, but sure

pkg/rehearse/jobs.go
droslean: Use Selector from [apimachinery/pkg/labels](https://godoc.org/k8s.io/apimachinery/pkg/labels#Selector) maybe?

bbguimaraes:
@droslean, @petr-muller, @stevekuznetsov: updated, PTAL again.

pkg/rehearse/jobs.go
```diff
@@ -188,18 +201,70 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+		}
+		logger.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
+		pjs.Insert(created.Name)
+	}
+	req, err := labels.NewRequirement(rehearseLabel, selection.Equals, []string{strconv.Itoa(prNumber)})
+	if err != nil {
+		return fmt.Errorf("failed to create label selector: %v", err)
+	}
+	selector := labels.NewSelector().Add(*req).String()
+	success, err := waitForJobs(pjs, selector, pjclient, logger)
```
petr-muller, line 221: Are we sure this won't deadlock occasionally? 

I think this is a similar pattern we fixed in ci-operator earlier. Here we first submit all jobs **then** create a watcher and wait for events, but what if a prowjob manages to finish really quick and emit events before the watcher is even set up?

pkg/rehearse/jobs.go
```diff
@@ -188,18 +201,70 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+		for {
+			event, ok := <-ch
+			if !ok {
+				break
+			}
+			pj, ok := event.Object.(*pjapi.ProwJob)
+			if !ok {
+				return false, fmt.Errorf("received a %T from watch", event.Object)
+			}
+			logger.WithFields(pjutil.ProwJobFields(pj)).WithFields(logrus.Fields{"state": pj.Status.State}).Info("processing ProwJob")
```
petr-muller: nit: the second `WithFields` can be a simpler https://godoc.org/github.com/sirupsen/logrus#WithField

pkg/rehearse/jobs_test.go
```diff
@@ -274,8 +279,31 @@ func TestExecuteJobs(t *testing.T) {
...
 		t.Run(tc.description, func(t *testing.T) {
-			fakeclient := fake.NewSimpleClientset().ProwV1().ProwJobs(testNamespace)
-			err := ExecuteJobs(tc.jobs, testPrNumber, testRepo, testRefs, testLogger, fakeclient)
+			fakecs := fake.NewSimpleClientset()
+			fakeclient := fakecs.ProwV1().ProwJobs(testNamespace)
+			watcher, err := fakeclient.Watch(metav1.ListOptions{})
+			if err != nil {
+				t.Fatalf("Failed to setup watch: %v", err)
+			}
+			fakecs.Fake.PrependWatchReactor("prowjobs", func(clientgo_testing.Action) (bool, watch.Interface, error) {
```
petr-muller, line 337: Oh, so **this** is how it's done... I basically re-implemented something similar for ciop tests, stupid me :(

pkg/rehearse/jobs.go
bbguimaraes: Here is what the documentation for the `ListOptions` struct that is passed to the `Watch` call says:

https://github.com/kubernetes/apimachinery/blob/fa6ddc151d63306b3540a37d910a07b181e4a474/pkg/apis/meta/v1/types.go#L344

```go
// When specified with a watch call, shows changes that occur after that particular version of a resource.
// Defaults to changes from the beginning of history.
// …
ResourceVersion string `json:"resourceVersion,omitempty" protobuf:"bytes,4,opt,name=resourceVersion"`
```

I have also observed this in practice: before the label filter was added, every time I ran the program it would retell the long Tale of the 5000 Prowjobs. Note that it includes finished jobs, and the `switch` statement in `waitForJobs` will handle those correctly.

All this — and the fact that it behaved as expected in all my tests — indicate that the code is correct, although I don't know for certain.

pkg/rehearse/jobs.go
bbguimaraes: Nice, updated.

pkg/rehearse/jobs_test.go
bbguimaraes: Have a look at the fake packages, there are many useful things like this in there.

bbguimaraes:
Updated.

Also changed `for { x, ok := <-someChannel; if !ok { break } }`to `for x := range someChannel {}` (TIL).

pkg/rehearse/jobs.go
petr-muller: Cool, thanks for explanation. I was reading into this a bit yesterday evening and from what I understood it really works like this.

petr-muller:
/hold

I'd like to check one more thing

petr-muller:
Maybe I'm missing something, but won't this break with `--dry-run`? We won't create any ProwJobs on the server but then wait until they change state?

Edit: After a quick test it seems it really hangs.

bbguimaraes:
It will "break" in the sense that "you will have to `ctrl-c` it" because there obviously won't be events for non-existing jobs. But I thought dry-run mode was meant for debug.

petr-muller:
Yeah dry-run is for debug but it should IMO mimic what would happen on happy path. We're currently running in dry-run mode in the `ci/prow/pj-rehearse` presubmit, that job would stop terminating.

bbguimaraes:
What do you think we should do in `waitForJobs` in dry-run mode, then?

petr-muller:
One option is to assume that all created jobs would happily succeed and just skip all waiting in the dry mode. This is probably cleaner but currently `ExecuteJobs` does not know if it's running dry or not so the boolean would need to be passed in.

Another option is to make the client wrapper return a "not actually created, running in dry" error when running in dry mode instead of pretending it created something. This would cause the set of jobs to be waited on to be empty, resolving the problem for the simple cost of some warnings being emited. Which could be fine.

bbguimaraes:
Added a `follow` argument to `ExecuteJobs`, set to `!o.dryRun`.

pkg/rehearse/jobs.go
```diff
@@ -29,34 +32,26 @@ const (
...
+			pj := action.(testing.CreateAction).GetObject().(*pjapi.ProwJob)
+			jobAsYAML, err := yaml.Marshal(pj)
+			if err != nil {
+				return true, nil, fmt.Errorf("failed to marshal job to YAML: %v", err)
+			}
+			fmt.Printf("%s\n", jobAsYAML[:10])
+			return false, nil, nil
+		})
+		return pjcset.ProwV1().ProwJobs(namespace), nil
+	}
```
bbguimaraes, line 49: From an earlier (failed) attempt at solving the problem, this has the benefit of not even requiring a kubeconfig in dry-run mode.

Unfortunately, it would require an unhealthy amount of hacks to fake the jobs transitioning from pending to a finished state.

pkg/rehearse/jobs.go
```diff
@@ -20,34 +28,30 @@ import (
...
 func NewProwJobClient(clusterConfig *rest.Config, namespace string, dry bool) (pj.ProwJobInterface, error) {
+	if dry {
+		pjcset := pjclientsetfake.NewSimpleClientset()
+		pjcset.Fake.PrependReactor("create", "prowjobs", func(action testing.Action) (bool, runtime.Object, error) {
+			pj := action.(testing.CreateAction).GetObject().(*pjapi.ProwJob)
+			jobAsYAML, err := yaml.Marshal(pj)
+			if err != nil {
+				return true, nil, fmt.Errorf("failed to marshal job to YAML: %v", err)
+			}
+			fmt.Printf("%s\n", jobAsYAML[:10])
```
stevekuznetsov: What's in those ten characters?

pkg/rehearse/jobs.go
```diff
@@ -192,18 +200,68 @@ func ExecuteJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
-			created, err := submitRehearsal(job, refs, logger, pjclient)
-			if err != nil {
-				logger.WithError(err).Warn("Failed to execute a rehearsal presubmit")
-			} else {
-				logger.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
-			}
-		}
-	} else {
+	if len(rehearsals) == 0 {
 		logger.Warn("No job rehearsals")
```
stevekuznetsov: Not a warning really

stevekuznetsov (commented):
Looks good!

pkg/rehearse/jobs.go
bbguimaraes: That's embarrassing… It was annoying to get two screens of yaml thrown at me every time I tested it =)

Removed.

pkg/rehearse/jobs.go
bbguimaraes: That has been in there from the beginning. We have extra output now that makes this unnecessary. Removed.

petr-muller:
/lgtm

unhold at will

PR #52, opened 2019-01-07
Init: Find which presubmits have to be executed.

droslean:
Compares the `presubmits` from the decorated repository and the master files.
Finds which of them shall be executed and creates a mapping of the repo to `presubmits` to execute.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -1,0 +1,50 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"log"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/diffs"
+)
+
+type options struct {
+	configPath    string
+	jobConfigPath string
+}
+
+func gatherOptions() options {
+	o := options{}
+	flag.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to Prow config.yaml")
+	flag.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
+	flag.Parse()
+	return o
+}
+
+func validateOptions(o options) error {
+	if len(o.jobConfigPath) == 0 {
+		return fmt.Errorf("empty --job-config-path")
+	}
+	return nil
+}
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(o)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	changedPresubmits, err := diffs.GetChangedPresubmits(o.configPath, o.jobConfigPath)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+	}
+
+	// just print them for now.
+	for repo, jobs := range changedPresubmits {
+		fmt.Println(repo)
+		for _, job := range jobs {
+			fmt.Println(job.Name)
+		}
+	}
+}
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -1,0 +1,50 @@
+package diffs
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+// GetChangedPresubmits returns a mapping of repo to presubmits to execute.
+func GetChangedPresubmits(configPath, jobConfigPath string) (map[string][]prowconfig.Presubmit, error) {
+	ret := make(map[string][]prowconfig.Presubmit)
+
+	prowMasterConfig, err := prowconfig.Load(configPath, jobConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load master's Prow config: %v", err)
+	}
+
+	prowPRConfig, err := prowconfig.Load("./cluster/ci/config/prow/config.yaml", "./ci-operator/jobs/")
+	if err != nil {
+		return nil, fmt.Errorf("failed to load PR's Prow config: %v", err)
+	}
+
+	masterJobs := getJobsByRepoAndName(prowMasterConfig.JobConfig.Presubmits)
+	for repo, jobs := range prowPRConfig.JobConfig.Presubmits {
+		for _, job := range jobs {
+			if !equality.Semantic.DeepEqual(masterJobs[repo][job.Name].Spec, job.Spec) {
+				ret[repo] = append(ret[repo], job)
+			}
+		}
+	}
+	return ret, nil
+}
+
+// To compare two maps of slices, instead of iterating through the slice
+// and compare the same key and index of the other map of slices,
+// we convert them as `repo-> jobName-> Presubmit` to be able to
+// access any specific elements of the Presubmits without the need to iterate in slices.
+func getJobsByRepoAndName(presubmits map[string][]prowconfig.Presubmit) map[string]map[string]prowconfig.Presubmit {
+	jobsByRepo := make(map[string]map[string]prowconfig.Presubmit)
+
+	for repo, preSubmitList := range presubmits {
+		pm := make(map[string]prowconfig.Presubmit)
+		for _, p := range preSubmitList {
+			pm[p.Name] = p
+		}
+		jobsByRepo[repo] = pm
+	}
+	return jobsByRepo
+}
--- a/images/pj-rehearse/Dockerfile
+++ b/images/pj-rehearse/Dockerfile
@@ -1,0 +1,5 @@
+FROM centos:7
+LABEL maintainer="nmoraiti@redhat.com"
+
+ADD pj-rehearse /usr/bin/pj-rehearse
+ENTRYPOINT ["/usr/bin/pj-rehearse"]


cmd/pj-rehearse/main.go
```diff
@@ -0,0 +1,53 @@
...
+	configPath     string
+	jobConfigPath  string
+	decorationPath string
+}
+
+func gatherOptions() options {
+	o := options{}
+	flag.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to Prow config.yaml")
+	flag.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to prow job configs.")
+	flag.StringVar(&o.decorationPath, "decoration-path", "./", "Path where the repository has been cloned.")
```
petr-muller: I am not that happy with the `decorationPath` name, as it does not really describe what it is. I think we should accept symmetrical options for master and under-test configs. Either accept paths to full `openshift/release` repo clones (one for master, one for PR) and compute the `jobConfig` paths by knowing the layout of `openshift/release`, or accept two sets of options (again, one for master, one for pr) pointing to needed directories directly.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,66 @@
...
+func NewDiffs(configPath, jobConfigPath, decorationPath string) *Diffs {
+	return &Diffs{
+		ConfigPath:     configPath,
+		JobConfigPath:  jobConfigPath,
+		DecorationPath: decorationPath,
+	}
+}
+
+// GetPresubmitsToExecute returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetPresubmitsToExecute() (map[string][]prowconfig.Presubmit, error) {
```
petr-muller: I think do not assume what the Presubmits are for. They are Presubmits affected by a proposed change, what they are used for is not this code's business. The ...ToExecute suffix increases the code coupling unnecessarily.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,66 @@
...
+	prowPRConfig, err := prowconfig.Load(fmt.Sprintf("%s/cluster/ci/config/prow/config.yaml", d.DecorationPath), fmt.Sprintf("%s/ci-operator/jobs/", d.DecorationPath))
+	if err != nil {
+		return nil, fmt.Errorf("failed to load PR's Prow config: %v", err)
+	}
+
+	masterJobs = getJobsByRepoAndName(prowMasterConfig.JobConfig.Presubmits)
+	prJobs = getJobsByRepoAndName(prowPRConfig.JobConfig.Presubmits)
+	for repo, jobMap := range prJobs {
+		for jobName, job := range jobMap {
+			if !equality.Semantic.DeepEqual(masterJobs[repo][jobName].Spec, job.Spec) {
```
petr-muller: I think this will crash on newly added jobs. Newly added job's jobName won't have an appropriate master presubmit item, and we will attempt to access such item's Spec.

Also, are we really sure we do not want to rehearse jobs changed in non-Spec fields? Are all these fields that harmless?

@droslean 's comment in the original #50:
> We will have to be more specific in which fields we need to execute the job. Here I only check the .spec.

What do you mean about being more specific?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,66 @@
...
+	return &Diffs{
+		ConfigPath:     configPath,
+		JobConfigPath:  jobConfigPath,
+		DecorationPath: decorationPath,
+	}
+}
+
+// GetPresubmitsToExecute returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetPresubmitsToExecute() (map[string][]prowconfig.Presubmit, error) {
+	preSubmitsToExecuteMap := make(map[string][]prowconfig.Presubmit)
```
petr-muller: Do we need to pollute the names with type information (the `Map` suffix)? It's rarely useful elsewhere but maybe I'm missing some Go folk knowledge.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,66 @@
...
+
+	var prJobs map[string]map[string]prowconfig.Presubmit
+	var masterJobs map[string]map[string]prowconfig.Presubmit
+
+	prowMasterConfig, err := prowconfig.Load(d.ConfigPath, d.JobConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load master's Prow config: %v", err)
+	}
+
+	prowPRConfig, err := prowconfig.Load(fmt.Sprintf("%s/cluster/ci/config/prow/config.yaml", d.DecorationPath), fmt.Sprintf("%s/ci-operator/jobs/", d.DecorationPath))
```
petr-muller: The path contatenation should be done with `filepath.Join` istead of `fmt.Sprintf`.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,66 @@
...
+		for jobName, job := range jobMap {
+			if !equality.Semantic.DeepEqual(masterJobs[repo][jobName].Spec, job.Spec) {
+				preSubmitsToExecuteMap[repo] = append(preSubmitsToExecuteMap[repo], job)
+			}
+		}
+	}
+	return preSubmitsToExecuteMap, nil
+}
+
+func getJobsByRepoAndName(presubmits map[string][]prowconfig.Presubmit) map[string]map[string]prowconfig.Presubmit {
```
petr-muller, line 39: Maybe a small comment explaining why this function needs to exist (along the lines of the original thread here https://github.com/openshift/ci-operator-prowgen/pull/50#discussion_r245327669) whould be helpful for the future readers.

petr-muller (commented):
I like this version. It's nice how it got really straightforward and simple. Let's iron out the code comments (inline, mostly taken from the original #50 PR) and let's get it merged so that we finally have a base to iterate on (I can then rebase my PR on this easily). I would even say let's merge it without tests and proivde them in a followup PR.

pkg/diffs/diffs.go
droslean: It won't crash, I also tested it with newly added repo as well.  What other filed we should consider worth to rehearse except the .spec?

pkg/diffs/diffs.go
droslean: with `make` I make sure that this map won't be nil. In that case, it will create an empty map.

droslean: Also, the reason why it won't crash mentioned in [#comment](https://github.com/openshift/ci-operator-prowgen/pull/52#discussion_r245669029)

cmd/pj-rehearse/main.go
```diff
@@ -0,0 +1,51 @@
+package main
```
bbguimaraes, line 1: It will be easier to integrate this if you use `cmd/rehearse/main.go` from Petr's PR.

cmd/pj-rehearse/main.go
```diff
@@ -0,0 +1,51 @@
...
+	o := gatherOptions()
+	err := validateOptions(o)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	diffs := diffs.NewDiffs(o.configPath, o.jobConfigPath)
+	preSubmitsToExecute, err := diffs.GetChangedPresubmits()
+	if err != nil {
+		fmt.Printf("Error: %v", err)
```
bbguimaraes: Missing a new-line character here (mentioning just in case it gets accidentally transplanted to the actual `main.go` file).

pkg/diffs/diffs.go
bbguimaraes: Following up from Petr's earlier suggestion, this variable should be named differently. Personally, I would choose `ret`: short and instantly conveys its purpose.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,71 @@
...
+		JobConfigPath: jobConfigPath,
+	}
+}
+
+// GetChangedPresubmits returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetChangedPresubmits() (map[string][]prowconfig.Presubmit, error) {
+	preSubmitsToExecuteMap := make(map[string][]prowconfig.Presubmit)
+
+	var prJobs map[string]map[string]prowconfig.Presubmit
+	var masterJobs map[string]map[string]prowconfig.Presubmit
```
bbguimaraes: No need for these declarations if the variables are not used until L41.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,71 @@
...
+		return nil, fmt.Errorf("failed to load master's Prow config: %v", err)
+	}
+
+	prowPRConfig, err := prowconfig.Load("./cluster/ci/config/prow/config.yaml", "./ci-operator/jobs/")
+	if err != nil {
+		return nil, fmt.Errorf("failed to load PR's Prow config: %v", err)
+	}
+
+	masterJobs = getJobsByRepoAndName(prowMasterConfig.JobConfig.Presubmits)
+	prJobs = getJobsByRepoAndName(prowPRConfig.JobConfig.Presubmits)
```
bbguimaraes: `prowPRConfig.JobConfig.Presubmits` can be used to drive the loop, there is no need to create a `map` for it.

pkg/diffs/diffs.go
bbguimaraes: Accessing non-existing keys will yield an uninitialized map followed by an uninitialized `ProwJobSpec`, which should be fine if `DeepEqual` can handle the later.

It may be faster to explicitly check for their existence, but the tool seemed fast enough in my tests.

pkg/diffs/diffs.go
bbguimaraes: The function can be inlined if we remove the unnecessary call above.

bbguimaraes (changes requested):
A couple of comments, mostly on implementation details.

@droslean, @petr-muller: how do we want to combine these PRs? I made mine a super-set of Petr's so they can be merged trivially via a fast-forward merge (before or after the merge to master).

petr-muller:
> how do we want to combine these PRs? I made mine a super-set of Petr's so they can be merged trivially via a fast-forward merge (before or after the merge to master).

I don't mind rebasing my PR (which up to now served more like an experiment, showcase vessel and not a real PR) on top of this. So I vote to resolve the little things here and merge it, IMO it's the easiest way.

pkg/diffs/diffs.go
petr-muller: Sorry, I spent too much time in my life using languages where uninitialized stuff bites you in the butt.

pkg/diffs/diffs.go
droslean: I can use it to iterate, but I can't use it to compare it with the master's `prowPRConfig.JobConfig.Presubmits`.

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+// Diffs holds the information where the differences can be found.
+type Diffs struct {
```
stevekuznetsov: What's the utility of this struct?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,68 @@
...
+	ConfigPath    string
+	JobConfigPath string
+}
+
+// NewDiffs creates a new Diffs{}
+func NewDiffs(configPath, jobConfigPath string) *Diffs {
```
stevekuznetsov: If they're exported fields, why bother with a constructor?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,68 @@
...
+// GetChangedPresubmits returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetChangedPresubmits() (map[string][]prowconfig.Presubmit, error) {
+	ret := make(map[string][]prowconfig.Presubmit)
+
+	prowMasterConfig, err := prowconfig.Load(d.ConfigPath, d.JobConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load master's Prow config: %v", err)
+	}
+
+	prowPRConfig, err := prowconfig.Load("./cluster/ci/config/prow/config.yaml", "./ci-operator/jobs/")
```
stevekuznetsov, line 19: Err - why are these hard-coded?

pkg/diffs/diffs.go
```diff
@@ -0,0 +1,68 @@
...
+// NewDiffs creates a new Diffs{}
+func NewDiffs(configPath, jobConfigPath string) *Diffs {
+	return &Diffs{
+		ConfigPath:    configPath,
+		JobConfigPath: jobConfigPath,
+	}
+}
+
+// GetChangedPresubmits returns a mapping of repo to presubmits to execute.
+func (d *Diffs) GetChangedPresubmits() (map[string][]prowconfig.Presubmit, error) {
```
stevekuznetsov: This is somehow expecting the PR to be checked out in a different location from the current working directory. In what set-up is this true? Will this be true in a PR job? Why have we chosen this?

pkg/diffs/diffs.go
stevekuznetsov: There are dozens of fields outside of spec. Why not diff on them all? If I change an `agent: kubernetes` job to `agent: jenkins` I would expect this rehearsal to catch it.

stevekuznetsov (commented):
I think this portion of the rehearsal should start from the working directory of a job - PR is merged into the target branch. Can we use `git` and other tools to determine what the config state was before and after the proposed changes?

pkg/diffs/diffs.go
droslean: We assume that those files are in the parent directory because of the decoration.

droslean:
@stevekuznetsov Answering to your [comment](https://github.com/openshift/ci-operator-prowgen/pull/52#discussion_r245847063): If we need to check all the fields of the job then we can't use `equality.Semantic.DeepEqual` because the struct contains unexporting fields. I think also that its better to pick the specific fields that we want to compare, to avoid reruning the jobs for changes in fields like `rerun_command` etc...

pkg/diffs/diffs.go
bbguimaraes:
```go
masterJobs := getJobsByRepoAndName(prowMasterConfig.JobConfig.Presubmits)
for repo, jobs := range prowPRConfig.JobConfig.Presubmits {
    for _, job := range jobs {
        if !equality.Semantic.DeepEqual(masterJobs[repo][job.Name].Spec, job.Spec) {
            preSubmitsToExecuteMap[repo] = append(preSubmitsToExecuteMap[repo], job)
        }
    }
}
```

droslean: good catch. updated

petr-muller:
We discussed this on Slack, here's the summary: the only remaining big question was the availability of both PR and master configs to the Prow job running the rehearse. PR version is available via decoration and master version is available in the cluster and it can be mounted to the Prowjob. We agreed we will merge this PR as a base for further development and set up full final `openshift/release` presubmit to run it (https://github.com/openshift/release/pull/2520, https://github.com/openshift/release/pull/2522). Right now that presubmit will only list the jobs that will be rehearsed in the future, which should be totally safe. We can incrementally develop it and this early deployment can keep us honest about not merging broken stuff :)

/lgtm
/hold for https://github.com/openshift/release/pull/2520 to merge so it can test this PR.

pkg/diffs/diffs.go
stevekuznetsov: Let's not do this.

PR #53, opened 2019-01-08
Bump the vendored deps to latest test-infra

petr-muller:
We will need the latest test-infra for the rehearsal work.

1. It needed an `[[override]]` for `Azure/go-autorest` to get rid of
some build problems unrelated to our code. I haven't really dig deep
into this and just found something promising on the internet and it
worked.
2. I needed to add a `[[constraint]]` on test-infra, without that `dep
ensure` crashed for me :/

The `k8s.io/test-infra/prow/config` structs used by Prowgen changed a
bit in the meantime (elements common for both `Presubmit` and
`Postsubmit` configs were extracted to a `JobBase` struct) which needed
some adjustments in Prowgen itself. Unfortunately the changes also
slightly changed the way how the job configs are marshalled in YAML, so
the update will also need a big `openshift/release` PR:

```
     decorate: true
+    decoration_config:
+      skip_cloning: true
     labels:
       artifacts: images
     name: branch-ci-super-duper-release-3.11-images
-    skip_cloning: true
```


petr-muller:
The corresponding PR for openshift/release: https://github.com/openshift/release/pull/2527

petr-muller:
/hold

Looks like I made a mistake somewhere, this seems to remove some code that it shouldn't

petr-muller:
/hold cancel

Now it's fine. I have updated https://github.com/openshift/release/pull/2527 too.

PR #55, opened 2019-01-18
Suppress rehearsing jobs with additional volumes

petr-muller:
Mounted volumes is currently the way how the templates are passed to
ci-operator in template-based jobs, generated by Prowgen. In these jobs,
volumes for templates are populated from the appropriate ConfigMap
present in the cluster. If we attempted to rehearse these jobs, they
would also use the content in the cluster. But the templates can be
also changed in the same PR as for which we are running the rehearsal,
which means a naive rehearsal would run over a different version of the
template than the one being present if the tested PR would be merged.
Therefore, do not rehearse these jobs for now, until we have a reliable
mechanism to pass all necessary content to rehearsed jobs.

An auxiliary benefit of this suppression is that it limits job
rehearsals only to the standard, container-based jobs which tend to be
simpler, making feature simpler and less prone to bugs, which is a good
thing for an initial release.

### DIFF
--- a/pkg/rehearse/jobs_test.go
+++ b/pkg/rehearse/jobs_test.go
@@ -166,6 +166,55 @@ func TestMakeRehearsalPresubmit(t *testing.T) {
 	}
 }
 
+func TestMakeRehearsalPresubmitNegative(t *testing.T) {
+	testName := "pull-ci-organization-repo-master-test"
+	testContext := "ci/prow/test"
+	testArgs := []string{"arg"}
+	testRepo := "organization/repo"
+	testPrNumber := 321
+
+	testCases := []struct {
+		description string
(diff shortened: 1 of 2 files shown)


pkg/rehearse/jobs_test.go
```diff
@@ -166,6 +166,55 @@ func TestMakeRehearsalPresubmit(t *testing.T) {
 	}
 }
 
+func TestMakeRehearsalPresubmitNegative(t *testing.T) {
```
stevekuznetsov, line 169: Love this

stevekuznetsov (commented):
+4 LOC and +49 lines of tests .... :love_letter: 

/lgtm

PR #56, opened 2019-01-22
Change from configMaps to master's worktree

droslean:
Since we have already the `base_rsa`, we add the master worktree and use that path to load the master's Prowconfig.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
+	"path"
 
 	"github.com/sirupsen/logrus"
 
@@ -54,9 +56,6 @@ func gatherOptions() options {
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 	fs.BoolVar(&o.noFail, "no-fail", true, "Whether to actually end unsuccessfuly when something breaks")
 
-	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to *master* Prow config.yaml")
-	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to *master* Prow Prow job configs.")
-
 	fs.StringVar(&o.candidatePath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
 
 	fs.Parse(os.Args[1:])
@@ -64,14 +63,9 @@ func gatherOptions() options {
 }
 
 func validateOptions(o options) error {
-	if len(o.jobConfigPath) == 0 {
-		return fmt.Errorf("--job-config-path was not provided")
-	}
-
 	if len(o.candidatePath) == 0 {
 		return fmt.Errorf("--candidate-path was not provided")
 	}
-
 	return nil
 }
 
@@ -111,7 +105,13 @@ func main() {
 
 	logger.Info("Rehearsing Prow jobs for a configuration PR")
 
-	prowConfig, err := prowconfig.Load(o.configPath, o.jobConfigPath)
+	masterWorktree, err := createMasterWorktree(o.candidatePath, jobSpec.Refs.BaseSHA)
+	if err != nil {
+		logger.WithError(err).Error("could not create master worktree")
+		gracefulExit(o.noFail)
+	}
+
+	prowConfig, err := prowconfig.Load(path.Join(masterWorktree, diffs.ConfigInRepoPath), path.Join(masterWorktree, diffs.JobConfigInRepoPath))
 	if err != nil {
 		logger.WithError(err).Error("Failed to load Prow config")
 		gracefulExit(o.noFail)
@@ -140,4 +140,13 @@ func main() {
 		logger.WithError(err).Error("Failed to execute rehearsal jobs")
 		gracefulExit(o.noFail)
 	}
+}
+
+func createMasterWorktree(candidatePath, baseSHA string) (string, error) {
+	cmd := exec.Command("git", "-C", candidatePath, "worktree", "add", "master", baseSHA)
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("'%s' failed with out: %s and error %v", cmd.Args, stdoutStderr, err)
+	}
+	return path.Join(candidatePath, "master"), nil
 }
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -54,27 +56,14 @@ func gatherOptions() options {
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 	fs.BoolVar(&o.noFail, "no-fail", true, "Whether to actually end unsuccessfuly when something breaks")
 
-	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to *master* Prow config.yaml")
-	fs.StringVar(&o.jobConfigPath, "job-config-path", "", "Path to *master* Prow Prow job configs.")
-
-	fs.StringVar(&o.candidatePath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
+	fs.StringVar(&o.configPath, "config-path", "master/cluster/ci/config/prow/config.yaml", "Path to *master* Prow config.yaml")
+	fs.StringVar(&o.jobConfigPath, "job-config-path", "master/ci-operator/jobs/", "Path to *master* Prow Prow job configs.")
+	fs.StringVar(&o.candidatePath, "candidate-path", ".", "Path to a openshift/release working copy with a revision to be tested")
```
petr-muller: Candidate not having a default was intentional, so I would prefer to keep it that way. See https://github.com/openshift/ci-operator-prowgen/pull/49#discussion_r246854375.

IMO, with master paths being inferred from candidatePath, we no longer need `configPath` and `jobConfigPath` at all: they can be inferred from wherever `createMasterWorktree` puts the worktree + the paths inside the repo we already know (they are not exported but we can make them):

https://github.com/openshift/ci-operator-prowgen/blob/684ed078be6bec78ee7a08ccd01afd8fe44d91a8/pkg/diffs/diffs.go#L11-L12

cmd/pj-rehearse/main.go
```diff
@@ -141,3 +130,13 @@ func main() {
 		gracefulExit(o.noFail)
 	}
 }
+
+func createMasterWorktree(baseRSA string) error {
+	var out bytes.Buffer
+	cmd := exec.Command("git", "worktree", "add", "master", baseRSA)
```
petr-muller: This assumes we run inside a clone and `o.candidatePath` was not provided leading somewhere else. `git` has a `-C` option that could be used here.

Also, I think this should explicitly return the path to the master worktree, so it can be used to infer paths to master config and master job configs.

petr-muller (changes requested):

bbguimaraes:
Is the local repository guaranteed to have the base branch? We can always pull it if it's not.

petr-muller:
> Is the local repository guaranteed to have the base branch? We can always pull it if it's not.

BaseSHA should be available in the local repo prepared for presubmits by clonerefs decoration, so not having it should happen only when you run locally e.g. for testing. We could fetch it if not available, but IMO it's not that important.

cmd/pj-rehearse/main.go
```diff
@@ -141,3 +142,13 @@ func main() {
 		gracefulExit(o.noFail)
 	}
 }
+
+func createMasterWorktree(cadidatePath, baseRSA string) error {
```
petr-muller: typo: cadidatePath -> candidatePath

cmd/pj-rehearse/main.go
```diff
@@ -111,7 +108,11 @@ func main() {
 
 	logger.Info("Rehearsing Prow jobs for a configuration PR")
 
-	prowConfig, err := prowconfig.Load(o.configPath, o.jobConfigPath)
+	if err := createMasterWorktree(o.candidatePath, jobSpec.Refs.BaseSHA); err != nil {
+		logger.WithError(err).Error("could not create master worktree")
+		gracefulExit(o.noFail)
+	}
+	prowConfig, err := prowconfig.Load(masterConfigPath, masterJobConfigPath)
```
petr-muller: This won't work if `o.candidatePath` is not `.`

The `master` directory containing the base will be created within the `o.candidatePath` directory.
`createMasterWorktree` should IMO return the master worktree path (`string.Join(o.candidatePath, "master")`) and we should create `master{Job}ConfigPath` using the paths in the `diffs` package.

petr-muller (changes requested):

cmd/pj-rehearse/main.go
stevekuznetsov: `path.Join`

cmd/pj-rehearse/main.go
```diff
@@ -141,3 +142,13 @@ func main() {
 		gracefulExit(o.noFail)
 	}
 }
+
+func createMasterWorktree(cadidatePath, baseRSA string) error {
+	var out bytes.Buffer
+	cmd := exec.Command("git", "-C", cadidatePath, "worktree", "add", "master", baseRSA)
+	cmd.Stdout = &out
```
stevekuznetsov: What is this used for? Could just use `CombinedOutput()` to get `stdout` and `stderr` and print on failures.

cmd/pj-rehearse/main.go
```diff
@@ -141,3 +142,13 @@ func main() {
 		gracefulExit(o.noFail)
 	}
 }
+
+func createMasterWorktree(cadidatePath, baseRSA string) error {
+	var out bytes.Buffer
+	cmd := exec.Command("git", "-C", cadidatePath, "worktree", "add", "master", baseRSA)
```
stevekuznetsov: This is a SHA not an RSA

cmd/pj-rehearse/main.go
petr-muller: doh, indeed :)

cmd/pj-rehearse/main.go
```diff
@@ -141,3 +145,12 @@ func main() {
 		gracefulExit(o.noFail)
 	}
 }
+
+func createMasterWorktree(candidatePath, baseSHA string) (string, error) {
+	cmd := exec.Command("git", "-C", candidatePath, "worktree", "add", "master", baseSHA)
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("'%s' failed with error %v", cmd.Args, err)
```
stevekuznetsov: We want to see the output on failure

cmd/pj-rehearse/main.go
```diff
@@ -111,7 +108,14 @@ func main() {
 
 	logger.Info("Rehearsing Prow jobs for a configuration PR")
 
-	prowConfig, err := prowconfig.Load(o.configPath, o.jobConfigPath)
+	stdoutStderr, err := createMasterWorktree(o.candidatePath, jobSpec.Refs.BaseSHA)
+	if err != nil {
+		logger.WithError(err).Error("could not create master worktree")
```
stevekuznetsov: Not sure there is a good reason to ferry around the stdout like this

cmd/pj-rehearse/main.go
```diff
@@ -19,6 +21,9 @@ import (
 	"github.com/openshift/ci-operator-prowgen/pkg/rehearse"
 )
 
+const masterConfigPath = "master/cluster/ci/config/prow/config.yaml"
+const masterJobConfigPath = "master/ci-operator/jobs/"
```
petr-muller: One of my comments was not addressed, and I'm not sure whether it's an omission or intentional (because it was not explicity refused), so I'll reiterate.

I don't think these constants are necessary and even desirable - we have the paths relative to `openshift/release` repo available already, in the `pkg/diffs` package. They are not exported, but that can change because the paths are useful outside. The only difference is the `master/` prefix, present because that is, accidentally, where `createMasterWorktree` creates the worktree. This means these constants are tightly coupled to `createMasterWorktree`, which is not desirable - the coupling is unclear, at minimum. 

We can easily get rid of these constants if `createMasterWorktree` returns a path where it created the worktree, which is IMO a reasonable API. Such returned path can then be combined with the constants from the `pkg/diffs` package.

petr-muller (commented):
One of my comments was not addressed, and I'm not sure whether it's an omission or intentional (because it was not explicity refused), so I'll reiterate

cmd/pj-rehearse/main.go
droslean: I misunderstood your previous [comment](https://github.com/openshift/ci-operator-prowgen/pull/56#discussion_r249768324). Changes addressed in the latest push.

stevekuznetsov (commented):
How can we test this?

droslean:
I locally tested, and it seems ok. To test it as a job, will be tricky. I have pushed the image in the `pj-rehearse-test` imagestream though

stevekuznetsov:
Let's think about how we will test this end-to-end automatically

droslean:
Shall we create a `presubmit` to release repo, with `always_run:false` `optional:true` and point the image to pj-rehearse-test imagestream? Just for temporary solution. And then we can use @petr-muller [test PR](2596) to rerun the job and test.

@stevekuznetsov @bbguimaraes @petr-muller  Any thoughts?

stevekuznetsov:
No, can we write an end-to-end test for this repo?

droslean:
Related PR in release repo https://github.com/openshift/release/pull/2649

droslean:
@stevekuznetsov I have tested this PR in our stage namespace. (see pod/cb8b9fa9-23af-11e9-93a7-54e1ad07d68a logs). I am currently working on creating some e2e tests for the repo. Also, I've opened #62 to satisfy the current changes. We need to merge #62 first.

bbguimaraes:
Nice.

/lgtm

stevekuznetsov:
Hrm, still would love to see a test of the functionality in this repo

PR #57, opened 2019-01-22
skip changed jobs with no branches.

droslean:
Skip all changed jobs without branches.


petr-muller (commented):
I'm slightly leaning towards keeping the `GetChangedPresubmits` simply returning all jobs that we *should* rehearse, and only later filtering out those that we cannot or want not rehearse. There is already some filtering-out logic like that in https://github.com/openshift/ci-operator-prowgen/blob/master/pkg/rehearse/jobs.go#L53 so I would prefer to concentrate it there, for separation of concerns.

PR #58, opened 2019-01-24
Allow `openshift-4.0` branches to be promoted

smarterclayton:
Discovered in https://github.com/openshift/ci-operator-prowgen/pull/58, etcd has to have `openshift-4.0` branch because upstream has `release-X`.

I need to come back to this and make this not be a hardcoded list here.


stevekuznetsov (approved):
/lgtm

PR #60, opened 2019-01-26
Better logging, job validation and unitests.

droslean:
Do a more detailed comparison of the jobs and output logs with the changes information.
Split the job validation logic and added unit tests for diff package.

WIP more unit tests.

Log output for the job differences examples:
```
INFO[0000] Job has been chosen for rehearsal                          diffs=" .Spec.Volumes[0].Name:   a: 'cluster-profile'   b: 'cluster-profile-test' .Spec.Volumes[0].VolumeSource.Projected.Sources[0].Secret.LocalObjectReference.Name:   a: 'cluster-secrets-aws'   b: 'clustser-secrets-aws'" job-name=pull-ci-openshift-cluster-autoscaler-operator-master-e2e-aws pr=2596 repo=openshift/cluster-autoscaler-operator
```
```
INFO[0000] Job has been chosen for rehearsal                          diffs=" .Spec.Containers[0].Args[0]:   a: '--artifact-dir=$(ARTIFACTS)'   b: '--artifact-dir=$(ASRTIFACTS)'" job-name=pull-ci-heketi-heketi-master-simple pr=2596 repo=heketi/heketi
```
```
INFO[0000] Job has been chosen for rehearsal                          diffs=" .Agent:   a: 'kubernetes'   b: 'jenkins'" job-name=pull-ci-openshift-router-master-e2e-aws pr=2596 repo=openshift/router
```

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -50,28 +50,6 @@ type options struct {
 	candidatePath string
 }
 
-func getMasterProwConfig(repoPath, sha, configPath, jobConfigPath string) (*prowconfig.Config, error) {
-	currentSHA, err := getCurrentSHA(repoPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get SHA of current HEAD: %v", err)
-	}
-
-	if err := gitCheckout(repoPath, sha); err != nil {
-		return nil, fmt.Errorf("failed to check out baseline revision: %v", err)
-	}
-
-	masterProwConfig, err := prowconfig.Load(configPath, jobConfigPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load baseline Prow config: %v", err)
-	}
-
-	if err := gitCheckout(repoPath, currentSHA); err != nil {
-		return nil, fmt.Errorf("failed to check out tested revision back: %v", err)
-	}
-
-	return masterProwConfig, nil
-}
-
 func gatherOptions() options {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
@@ -96,7 +74,6 @@ func gracefulExit(suppressFailures bool) {
 	if suppressFailures {
 		os.Exit(0)
 	}
-
 	os.Exit(1)
 }
 
@@ -125,25 +102,7 @@ func main() {
 
 	prNumber := jobSpec.Refs.Pulls[0].Number
 	logger = logrus.WithField(prowgithub.PrLogField, prNumber)
-
 	logger.Info("Rehearsing Prow jobs for a configuration PR")
-
-	configPath := filepath.Join(o.candidatePath, diffs.ConfigInRepoPath)
-	jobConfigPath := filepath.Join(o.candidatePath, diffs.JobConfigInRepoPath)
-
-	prowPRConfig, err := prowconfig.Load(configPath, jobConfigPath)
-	if err != nil {
-		logger.WithError(err).Error("Failed to load PR's Prow config")
-		gracefulExit(o.noFail)
-	}
-
-	masterProwConfig, err := getMasterProwConfig(o.candidatePath, jobSpec.Refs.BaseSHA, configPath, jobConfigPath)
-	if err != nil {
-		logger.WithError(err).Error("Failed to load master Prow config")
-		gracefulExit(o.noFail)
-	}
-
-	prowjobNamespace := masterProwConfig.ProwJobNamespace
 
 	var clusterConfig *rest.Config
 	if !o.dryRun {
@@ -154,13 +113,19 @@ func main() {
 		}
 	}
 
-	pjclient, err := rehearse.NewProwJobClient(clusterConfig, prowjobNamespace, o.dryRun)
+	prowConfig, prowPRConfig, err := getProwConfigs(o.candidatePath, jobSpec.Refs.BaseSHA)
+	if err != nil {
+		logger.WithError(err).Error("could not load prow configs")
+		gracefulExit(o.noFail)
+	}
+
+	pjclient, err := rehearse.NewProwJobClient(clusterConfig, prowConfig.ProwJobNamespace, o.dryRun)
 	if err != nil {
 		logger.WithError(err).Error("could not create a ProwJob client")
 		gracefulExit(o.noFail)
 	}
 
-	changedPresubmits, err := diffs.GetChangedPresubmits(masterProwConfig, prowPRConfig)
+	changedPresubmits, err := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
 	if err != nil {
 		logger.WithError(err).Error("Failed to determine which jobs should be rehearsed")
 		gracefulExit(o.noFail)
@@ -191,4 +156,34 @@ func gitCheckout(candidatePath, baseSHA string) error {
 		return fmt.Errorf("'%s' failed with out: %s and error %v", cmd.Args, stdoutStderr, err)
 	}
 	return nil
+}
+
+func getProwConfigs(candidatePath, baseSHA string) (*prowconfig.Config, *prowconfig.Config, error) {
+	currentSHA, err := getCurrentSHA(candidatePath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get SHA of current HEAD: %v", err)
+	}
+
+	candidateConfigPath := filepath.Join(candidatePath, diffs.ConfigInRepoPath)
+	candidateJobConfigPath := filepath.Join(candidatePath, diffs.JobConfigInRepoPath)
+
+	prowPRConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to load PR's Prow config: %v", err)
+	}
+
+	if err := gitCheckout(candidatePath, baseSHA); err != nil {
+		return nil, nil, fmt.Errorf("could not checkout worktree: %v", err)
+	}
+
+	masterProwConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to load master's Prow config: %v", err)
+	}
+
+	if err := gitCheckout(candidatePath, currentSHA); err != nil {
+		return nil, nil, fmt.Errorf("failed to check out tested revision back: %v", err)
+	}
+
+	return masterProwConfig, prowPRConfig, nil
 }
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -1,22 +1,57 @@
 package diffs
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
 	"k8s.io/apimachinery/pkg/api/equality"
+	utildiff "k8s.io/apimachinery/pkg/util/diff"
+	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 )
 
-const ConfigInRepoPath = "cluster/ci/config/prow/config.yaml"
-const JobConfigInRepoPath = "ci-operator/jobs"
+const (
+	// ConfigInRepoPath is the prow config path from release repo
+	ConfigInRepoPath = "cluster/ci/config/prow/config.yaml"
+	// JobConfigInRepoPath is the prowjobs path from release repo
+	JobConfigInRepoPath = "ci-operator/jobs"
+
+	logRepo    = "repo"
+	logJobName = "job-name"
+	logDiffs   = "diffs"
+
+	objectSpec  = ".Spec"
+	objectAgent = ".Agent"
+
+	chosenJob = "Job has been chosen for rehearsal"
+)
 
 // GetChangedPresubmits returns a mapping of repo to presubmits to execute.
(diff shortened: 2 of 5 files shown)


pkg/diffs/diffs.go
```diff
@@ -2,30 +2,68 @@ package diffs
...
 	"k8s.io/apimachinery/pkg/api/equality"
+	utildiff "k8s.io/apimachinery/pkg/util/diff"
 	prowconfig "k8s.io/test-infra/prow/config"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/rehearse"
 )
 
-const configInRepoPath = "cluster/ci/config/prow/config.yaml"
-const jobConfigInRepoPath = "ci-operator/jobs"
+const (
```
stevekuznetsov, line 15: Can we get some comments on these

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,68 @@ package diffs
...
+			if masterJob.Agent != job.Agent {
+				logFields[logDiffs] = convertToReadableDiff(masterJob.Agent, job.Agent, objectAgent)
+				logger.WithFields(logFields).Info(chosenJob)
+				ret[repo] = append(ret[repo], job)
+				continue
+			}
+
+			switch job.Agent {
+			case rehearse.JenkinsAgent:
+				// TODO: safely compare jenkins jobs
```
stevekuznetsov: I think we can always skip it

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,68 @@ package diffs
...
+
+			case rehearse.KubeAgent:
+				if !equality.Semantic.DeepEqual(masterJob.Spec.Containers, job.Spec.Containers) {
```
stevekuznetsov: Why can't we just compare the entire `Spec`?

pkg/diffs/diffs_test.go
```diff
@@ -0,0 +1,113 @@
...
+func TestGetChangedPresubmits(t *testing.T) {
+	minRequiredSpec := &v1.PodSpec{
+		Containers: []v1.Container{{
+			Image:   "test-image",
+			Command: []string{"ci-operator"},
+			Args:    []string{"--artifact-dir=$(ARTIFACTS)", "--target=images"},
+		}},
+	}
+
+	testCases := []struct {
```
stevekuznetsov, line 32: The intent of these tests is to change something and verify that the change ends up requiring the job, so perhaps we can refactor the test cases to be more succinct and easier to read:

```go
basePresubmit := &prowconfig.Presubmit{ /*....*/ }
testCases := []struct {
    name string
    configGenerator  func() (before, after *prowconfig.Config)
    expected map[string][]prowconfig.Presubmit
}{
    {
        name: "no differences mean nothing is identified as a diff",
        configGenerator: func() (before, after *prowconfig.Config) { return makeConfig(*basePresubmit), makeConfig(*basePresubmit) },
        expected: map[string][]prowconfig.Presubmit{},
    },

    {
        name: "different agent is identified as a diff",
        configGenerator: func() (before, after *prowconfig.Config) {
            after := *basePresubmit
            after.Agent = "foobar"
            return makeConfig(*basePresubmit), makeConfig(after)
        },
        expected: map[string][]prowconfig.Presubmit{ /* ... */ },
    },
}
```

pkg/rehearse/jobs.go
```diff
@@ -16,20 +16,23 @@ import (
...
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/pjutil"
-
-	"k8s.io/client-go/rest"
 )
 
 const (
 	rehearseLabel = "ci.openshift.org/rehearse"
+
+	KubeAgent    = "kubernetes"
```
stevekuznetsov: These should be imported

pkg/rehearse/jobs.go
```diff
@@ -65,38 +72,45 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
+		rehearsal.Spec.Containers[0].Args = append(source.Spec.Containers[0].Args, gitrefArg)
 	}
-	container := source.Spec.Containers[0]
+	return &rehearsal, nil
+}
 
-	if len(container.Command) != 1 || container.Command[0] != "ci-operator" {
-		return nil, fmt.Errorf("cannot rehearse jobs that have Command different from simple 'ci-operator'")
-	}
+func validateJob(source *prowconfig.Presubmit) error {
```
stevekuznetsov: This should be imported from upstream

stevekuznetsov: Err, this is not really validation but filtering -- if you want to validate that it's a good job, import. Rename the other checks to be filtersz

droslean: well, the validation is already happening when we load the prowconfig. So renaming makes sense.

pkg/rehearse/jobs.go
droslean: I can't compare it if I will import this. Is this intentional or a bug? https://github.com/kubernetes/test-infra/blob/master/prow/apis/prowjobs/v1/types.go#L67-L74

pkg/diffs/diffs.go
droslean: IMHO I would prefer at least output a log to the user and tell him that we don't rehearse it.  Or we can just skip it and include this information in our docs.

pkg/rehearse/jobs.go
stevekuznetsov: It should be comparable, if not, cast it to a ProwJobType then comapre

pkg/diffs/diffs.go
stevekuznetsov: Outputting a log sounds like a good idea to me

pkg/rehearse/jobs.go
droslean: It's not comparable and I cannot cast it without creating an interface for that, which will lead to a panic. I will keep looking for a better way, maybe I missed something here

droslean: my bad. It seems that `source.Agent == string(pjapi.KubernetesAgent)` is comparable. No need to cast it to a `ProwJobType`

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,69 @@ package diffs
...
+
+	objectSpec  = ".Spec"
+	objectAgent = ".Agent"
+
+	chosenJob = "Job has been chosen for rehearsal"
+)
 
 // GetChangedPresubmits returns a mapping of repo to presubmits to execute.
-func GetChangedPresubmits(prowMasterConfig *prowconfig.Config, candidateRepoPath string) (map[string][]prowconfig.Presubmit, error) {
+func GetChangedPresubmits(prowMasterConfig, prowPRConfig *prowconfig.Config, logger logrus.FieldLogger) (map[string][]prowconfig.Presubmit, error) {
```
stevekuznetsov: I think we usually pass `*logrus.Entry`, why `FieldLogger`?

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,69 @@ package diffs
...
+				logger.WithFields(logFields).Info(chosenJob)
+				ret[repo] = append(ret[repo], job)
+				continue
+			}
+
+			switch job.Agent {
+			case pjapi.JenkinsAgent:
+				job.SourcePath = masterJob.SourcePath
+				if !equality.Semantic.DeepEqual(masterJob.JobBase, job.JobBase) {
+					logFields[logDiffs] = convertToReadableDiff(masterJob.JobBase, job.JobBase, "")
```
stevekuznetsov: Not worth doing this diff

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,69 @@ package diffs
...
 	masterJobs := getJobsByRepoAndName(prowMasterConfig.JobConfig.Presubmits)
 	for repo, jobs := range prowPRConfig.JobConfig.Presubmits {
 		for _, job := range jobs {
-			if !equality.Semantic.DeepEqual(masterJobs[repo][job.Name].Spec, job.Spec) {
+			masterJob := masterJobs[repo][job.Name]
+			logFields := logrus.Fields{logRepo: repo, logJobName: job.Name}
+			chosen := false
+
+			// If the agent was changed and is a kubernetes agent, just choose the job for rehearse.
+			if masterJob.Agent != job.Agent && job.Agent == string(pjapi.KubernetesAgent) {
```
stevekuznetsov: Can we do this in the `switch`?

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,69 @@ package diffs
...
+				logFields[logDiffs] = convertToReadableDiff(masterJob.Agent, job.Agent, objectAgent)
+				logger.WithFields(logFields).Info(chosenJob)
+				ret[repo] = append(ret[repo], job)
+				continue
+			}
+
+			switch job.Agent {
```
stevekuznetsov: do a `switch pjapi.Agent(job.Agent)`?

pkg/diffs/diffs.go
```diff
@@ -2,30 +2,69 @@ package diffs
...
+			case pjapi.JenkinsAgent:
+				job.SourcePath = masterJob.SourcePath
+				if !equality.Semantic.DeepEqual(masterJob.JobBase, job.JobBase) {
+					logFields[logDiffs] = convertToReadableDiff(masterJob.JobBase, job.JobBase, "")
+					logger.WithFields(logFields).Warn("We don't rehearse jenkins jobs, skiping...")
+					continue
```
stevekuznetsov, line 47: Either do this or use the `chosen` bool, but both are doing the same thing

pkg/diffs/diffs_test.go
```diff
@@ -0,0 +1,132 @@
...
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+const orgRepo = "org/repo"
+
+func TestGetChangedPresubmits(t *testing.T) {
+	basePreSubmit := makeBasePresubmit()
+	testCases := []struct {
+		name                  string
+		masterConfigGenerator func() (config *prowconfig.Config)
```
stevekuznetsov: I think one func is cleaner -- you're always going to be taking A and changing it to get B, with two you must enforce that they both start with A

pkg/diffs/diffs_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+const orgRepo = "org/repo"
+
+func TestGetChangedPresubmits(t *testing.T) {
+	basePreSubmit := makeBasePresubmit()
```
stevekuznetsov: Just make this a variable in the test func, no need to make it a call

pkg/diffs/diffs_test.go
```diff
@@ -0,0 +1,132 @@
...
+import (
+	"testing"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+const orgRepo = "org/repo"
```
stevekuznetsov: Add this to the literal config, it's hard to thread through

pkg/diffs/diffs_test.go
```diff
@@ -0,0 +1,132 @@
...
+							VolumeSource: v1.VolumeSource{
+								EmptyDir: &v1.EmptyDirVolumeSource{},
+							}},
+						}
+						return *p
+					}()}},
+		},
+	}
+
+	for _, testCase := range testCases {
```
stevekuznetsov, line 133: Would like @petr-muller review on this before merge

pkg/rehearse/jobs_test.go
```diff
@@ -181,11 +183,6 @@ func TestMakeRehearsalPresubmitNegative(t *testing.T) {
 		description string
 		crippleFunc func(*prowconfig.Presubmit)
 	}{{
-		description: "job with multiple containers",
```
stevekuznetsov, line 185: Why?

droslean: this validation is already happening when we load the prowconfigs.

cmd/pj-rehearse/main.go
```diff
@@ -195,10 +176,14 @@ func getProwConfigs(candidatePath, baseSHA string) (*prowconfig.Config, *prowcon
 		return nil, nil, fmt.Errorf("could not checkout worktree: %v", err)
 	}
 
-	prowConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
+	masterRrowConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
```
petr-muller: typo:
```
masterRrowConfig
      ^
```

pkg/diffs/diffs_test.go
petr-muller: One testcase that I think is useful and we don't have it yet is the new job appearing.

petr-muller: Also, I started to like subtests in testsuites (https://blog.golang.org/subtests) - I find the output significantly clearer on failures, can we use them here too?

stevekuznetsov: Subtests should be the way forward for all go unit tests now -- others are just legacy

stevekuznetsov: They'll run faster, too

droslean: addressed

PR #62, opened 2019-01-29
add git to pj-rehearse image

droslean:


bbguimaraes:
If that is the only requirement, we can change the working directory in `rehearse` before `exec`ing `git`.

bbguimaraes:
As discussed in private chat, changing the base image or installing git from some random RPM server seems counterproductive to me in this case.

We can easily achieve what the convenience commands/flags we initially proposed give us without requiring a new version of `git`:

```go
cmd := exec.Command("git", "-C", path, …)
// can be exchanged with
cmd := exec.Command("git", …)
cmd.Dir = path
```

```go
git worktree add "$path" "$ref"
# can be exchanged with
git clone --branch "$ref" . "$path"
```

PR #65, opened 2019-01-30
Remove assumption that tag_specification is set

### LINKED ISSUE #61: panic when generating from a config that doesn't have a tag_specification and is trying to add an e2e test
I added an openshift_installer_src to the Hive configuration (https://github.com/openshift/release/pull/2719), and tried to generate prow jobs from it. However ci-operator-prowgen panics at: https://github.com/openshift/ci-operator-prowgen/blob/master/cmd/ci-operator-prowgen/main.go#L383 most likely because there is no ReleaseTagConfiguration in my config.

Should the config have a tag_specification? and where should it point to?

bbguimaraes:
Prevents a segfault when a configuration (correctly) omits
`tag_specification`.

Fixes #61.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -375,7 +375,11 @@ func generateJobs(
 		if element.ContainerTestConfiguration != nil {
 			podSpec = generatePodSpec(repoInfo.configFilename, element.As)
 		} else {
-			podSpec = generatePodSpecTemplate(repoInfo.org, repoInfo.repo, repoInfo.configFilename, configSpec.ReleaseTagConfiguration.Name, &element)
+			var release string
+			if c := configSpec.ReleaseTagConfiguration; c != nil {
+				release = c.Name
+			}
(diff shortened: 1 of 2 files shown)


petr-muller (commented):
I'm wondering whether we should also add a reasonable error check to the place where we would end up using the empty string as a part of the URL for the bona fide invalid configs.

bbguimaraes:
The sad thing about this bug is that everything else already handles this correctly. The configuration validation (part of `ci-operator`) already rejects invalid configuration files with a very specific error message.

The *only* problem here is we were accessing the `Name` member when we shouldn't: the empty value will never be used inside `generatePodSpecTemplate`.

bbguimaraes:
I.e. we had a properly protected castle, but I accidentally installed a guillotine instead of a door.

petr-muller:
Sounds good. I was afraid whether we couldn't merge some broken jobs mistakenly but we have a presubmit that checks checks for broken config, and another presumit that checks whether jobs were generated from most recent configs, and together it's pretty much all we need.

/lgtm

PR #67, opened 2019-02-04
pj-rehearse: tests for main, using integration data

droslean:
Add the basic integration flow test for `main` using also the integration data files.

### DIFF
--- a/cmd/pj-rehearse/main_test.go
+++ b/cmd/pj-rehearse/main_test.go
@@ -1,0 +1,151 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path/filepath"
+	"testing"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/watch"
+
+	clienttesting "k8s.io/client-go/testing"
+
+	"k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/client/clientset/versioned/fake"
+	prowconfig "k8s.io/test-infra/prow/config"
+	prowgithub "k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/kube"
+	pjdwapi "k8s.io/test-infra/prow/pod-utils/downwardapi"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/diffs"
+	"github.com/openshift/ci-operator-prowgen/pkg/rehearse"
+)
+
+const (
+	candidatePath    = "../../test/pj-rehearse-integration/candidate"
+	masterPath       = "../../test/pj-rehearse-integration/master"
+	expectedJobsFile = "../../test/pj-rehearse-integration/expected_jobs.yaml"
+)
+
+func TestIntegration(t *testing.T) {
+	jobSpec := pjdwapi.JobSpec{
+		Type:      kube.PresubmitJob,
+		Job:       "pull-pj-rehearse-demo",
+		BuildID:   "0",
+		ProwJobID: "prowjob",
+		Refs: &kube.Refs{
+			Org:     "super",
+			Repo:    "duper",
+			BaseRef: "base-ref",
+			BaseSHA: "base-sha",
+			Pulls: []v1.Pull{{
+				Number: 1234,
+				Author: "dptp-team",
+				SHA:    "a4dc6278a189aa66ae9c2a0f11834499f104fcd4",
+			}},
+		},
+	}
+
+	logger := logrus.WithFields(logrus.Fields{prowgithub.OrgLogField: jobSpec.Refs.Org, prowgithub.RepoLogField: jobSpec.Refs.Repo})
+	loggers := rehearse.Loggers{Job: logger, Debug: logger}
+	fakecs := fake.NewSimpleClientset()
+	fakeclient := fakecs.ProwV1().ProwJobs("test-namespace")
+
+	changedPresubmits, err := getRehersalsHelper(logger, jobSpec.Refs.Pulls[0].Number)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	watcher, err := fakeclient.Watch(metav1.ListOptions{})
+	if err != nil {
+		t.Fatalf("Failed to setup watch: %v", err)
+	}
+
+	toBeRehearsed := getExpectedProwJobs(t)
+	rehearsed := sets.NewString()
+
+	fakecs.Fake.PrependWatchReactor("prowjobs", func(clienttesting.Action) (bool, watch.Interface, error) {
+		watcher.Stop()
+		ret := watch.NewFakeWithChanSize(toBeRehearsed.Len(), true)
+
+		for event := range watcher.ResultChan() {
+			pj := event.Object.(*v1.ProwJob).DeepCopy()
+
+			if !toBeRehearsed.Has(pj.Spec.Job) {
+				t.Fatalf("Job %s shouldn't be rehearsed", pj.Spec.Job)
+			}
+
+			if rehearsed.Has(pj.Spec.Job) {
+				t.Fatalf("Job %s has already been rehearsed", pj.Spec.Job)
+			}
+
+			pj.Status.State = v1.SuccessState
+			ret.Modify(pj)
+
+			rehearsed.Insert(pj.Spec.Job)
+		}
+		return true, ret, nil
+	})
+
+	executor := rehearse.NewExecutor(changedPresubmits, jobSpec.Refs.Pulls[0].Number, candidatePath, jobSpec.Refs, false, loggers, fakeclient)
+	if _, err := executor.ExecuteJobs(); err != nil {
+		t.Fatalf("Failed to execute rehearsal jobs: %v", err)
+	}
+
+	if missing := rehearsed.Difference(toBeRehearsed); missing.Len() > 0 {
+		t.Fatalf("did not rehearse expected prowJobs: %v", missing.List())
+	}
+	if extra := toBeRehearsed.Difference(rehearsed); extra.Len() > 0 {
+		t.Fatalf("found unexpectedly rehearsed prowJobs: %v", extra.List())
+	}
+}
+
+func getExpectedProwJobs(t *testing.T) sets.String {
+	var expectedJobs []v1.ProwJob
+	expectedJobsNames := sets.NewString()
+
+	content, err := ioutil.ReadFile(expectedJobsFile)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = yaml.Unmarshal(content, &expectedJobs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, job := range expectedJobs {
+		expectedJobsNames.Insert(job.Spec.Job)
+	}
+
+	return expectedJobsNames
+}
+
+func getRehersalsHelper(logger *logrus.Entry, prNumber int) ([]*prowconfig.Presubmit, error) {
+	candidateConfigPath := filepath.Join(candidatePath, diffs.ConfigInRepoPath)
+	candidateJobConfigPath := filepath.Join(candidatePath, diffs.JobConfigInRepoPath)
+	masterConfigPath := filepath.Join(masterPath, diffs.ConfigInRepoPath)
+	masterJobConfigPath := filepath.Join(masterPath, diffs.JobConfigInRepoPath)
+
+	prowConfig, err := prowconfig.Load(masterConfigPath, masterJobConfigPath)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+func pjStatusChanger(fakecs *fake.Clientset, toBeRehearsed map[string][]prowconfig.Presubmit, refs *v1.Refs, t *testing.T) error {
+	fakeclient := fakecs.ProwV1().ProwJobs("test-namespace")
+
+	watcher, err := fakeclient.Watch(metav1.ListOptions{})
+	if err != nil {
+		t.Fatalf("Failed to setup watch: %v", err)
+	}
+
+	fakecs.Fake.PrependWatchReactor("prowjobs", func(clientgo_testing.Action) (bool, watch.Interface, error) {
```
stevekuznetsov: I think this import is most commonly `clienttesting`

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+				Author: "dptp-team",
+				SHA:    "a4dc6278a189aa66ae9c2a0f11834499f104fcd4",
+			}},
+		},
+	}
+
+	prFields := logrus.Fields{prowgithub.OrgLogField: jobSpec.Refs.Org, prowgithub.RepoLogField: jobSpec.Refs.Repo}
+	logger := logrus.WithFields(prFields)
+
+	candidateConfigPath := filepath.Join(candidatePath, diffs.ConfigInRepoPath)
```
stevekuznetsov: Can all this setup be called from shared code? This should not be local to the test

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	prowPRConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
+	if err != nil {
+		t.Fatalf("failed to load Prow config: %v", err)
+	}
+
+	fakecs := fake.NewSimpleClientset()
+	fakeclient := fakecs.ProwV1().ProwJobs("test-namespace")
+
+	changedPresubmits, err := diffs.GetChangedPresubmits(prowConfig, prowPRConfig)
```
stevekuznetsov: The manner which we determine changes and do something with them should also not be local to the test if we want to be more certain of the behavior of the binary itself

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+		t.Fatalf("Failed to determine which jobs should be rehearsed: %v", err)
+	}
+
+	pjStatusChanger(fakecs, changedPresubmits, jobSpec.Refs, t)
+	if err := rehearse.ExecuteJobs(changedPresubmits, jobSpec.Refs.Pulls[0].Number, candidatePath, jobSpec.Refs, true, logger, fakeclient); err != nil {
+		t.Fatalf("Failed to execute rehearsal jobs: %v", err)
+	}
+}
+
+func pjStatusChanger(fakecs *fake.Clientset, toBeRehearsed map[string][]prowconfig.Presubmit, refs *v1.Refs, t *testing.T) error {
```
stevekuznetsov: Why is this a function by itself?

cmd/pj-rehearse/main_test.go
stevekuznetsov: I don't really understand what this is doing -- what's the intent here?

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,103 @@
...
+	fakeclient := fakecs.ProwV1().ProwJobs("test-namespace")
+
+	changedPresubmits, err := diffs.GetChangedPresubmits(prowConfig, prowPRConfig)
+	if err != nil {
+		t.Fatalf("Failed to determine which jobs should be rehearsed: %v", err)
+	}
+
+	pjStatusChanger(fakecs, changedPresubmits, jobSpec.Refs, t)
+	if err := rehearse.ExecuteJobs(changedPresubmits, jobSpec.Refs.Pulls[0].Number, candidatePath, jobSpec.Refs, true, logger, fakeclient); err != nil {
+		t.Fatalf("Failed to execute rehearsal jobs: %v", err)
```
stevekuznetsov, line 97: Do we want to check anything other than "we didn't crash"? Or is that still WIP

stevekuznetsov (commented):
Looking sweet!

cmd/pj-rehearse/main_test.go
droslean: its still WIP, but I feel like to check only "we didn't crash" should be enough? We are already doing other tests to make sure that each step functions as it should be. Any other suggestions?

cmd/pj-rehearse/main_test.go
droslean: It will simulate a status change of the prowjob. Otherwise, `ExecuteJobs` will just hang forever.

droslean: asking in a state of ignorace, Is there any better way to achieve this?

petr-muller:
I will throw an opinion here, but it's more a topic for discussion than a straightforward suggestion:

Looking at these tests, the *actually tested* behavior is that of `ExecuteJobs` - but we already have tests of that function over a mocked cluster. The difference here is that these tests do not run it over the pre-prepared structrures (like the existing tests, being proper unit tests, do) but basically reimplement what `pj-rehearse` does in it's `main()`: prep some helper structs, parse some YAML, compare the resulting structs and call `ExecuteJobs` on it. This of course adds some testing value, but only as long as we do not change the `pj-rehearse`'s `main()` function.

What I *think* would be useful would be a full e2e test - prep a very little git repo resembling how the repo prepared by `cloneref` would look like, prep a `$JOB_SPEC` for it and run full `pj-rehearse` in dry mode over that. Non-dry behavior of `ExecuteJobs` is already unit-tested, so we should not be losing coverage by running in dry mode. We would validate `pj-rehearse` terminates successfuly, plus dry mode outputs rehearsal `ProwJob` YAMLs, which could be parsed and validated, if needed. We could only do rough validation on this level, though, like checking the count of jobs and their names - the details are already tested by unit tests.

Just a thought, though. I really do not want to imply this PR needs to be scrapped and rebuilt from the ground up, I'm just thinking about what the integration tests could test over what we test already.

droslean:
With this test, we will just make sure that the whole pipeline is behaving the way it should. I agree that e2e tests will be useful but I would like to have them both.

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	prowPRConfig, err := prowconfig.Load(candidateConfigPath, candidateJobConfigPath)
+	if err != nil {
+		t.Fatalf("failed to load Prow config: %v", err)
+	}
+
+	fakecs := fake.NewSimpleClientset()
+	fakeclient := fakecs.ProwV1().ProwJobs("test-namespace")
+
+	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
```
stevekuznetsov: Can we have a helper that takes the two paths as input and returns the diff, which we share between this test and the actual production code?

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,90 @@
...
+		ret := watch.NewFakeWithChanSize(size, true)
+		for event := range watcher.ResultChan() {
+			pj := event.Object.(*v1.ProwJob).DeepCopy()
+			pj.Status.State = v1.SuccessState
+			ret.Modify(pj)
+		}
+		return true, ret, nil
+	})
+
+	executor := rehearse.NewExecutor(changedPresubmits, jobSpec.Refs.Pulls[0].Number, candidatePath, jobSpec.Refs, true, loggers, fakeclient)
```
stevekuznetsov: I think all that's left is checking that we created the correct prowjobs, right?

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,130 @@
...
+				SHA:    "a4dc6278a189aa66ae9c2a0f11834499f104fcd4",
+			}},
+		},
+	}
+
+	fakecs     = fake.NewSimpleClientset()
+	fakeclient = fakecs.ProwV1().ProwJobs("test-namespace")
+)
+
+func TestIntergration(t *testing.T) {
```
petr-muller: Nit:
```
TestIntergration
        ^
```

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,130 @@
...
+		t.Fatal(err)
+	}
+
+	err = yaml.Unmarshal(content, &expectedJobs)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	for _, job := range expectedJobs {
+		expectedJobsNames = append(expectedJobsNames, job.Spec.Job)
```
petr-muller: Nit: maybe construct the `sets.String` right away and not `[]string` first

petr-muller (commented):
Few nits I'd like to see fixed but fine otherwise.

It will need some changes as we decide to refactor the code but I think it's worth merging this and then we can drive the refactor with test changes.

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"github.com/openshift/ci-operator-prowgen/pkg/rehearse"
+)
+
+const (
+	candidatePath    = "../../test/pj-rehearse-integration/candidate"
+	masterPath       = "../../test/pj-rehearse-integration/master"
+	expectedJobsFile = "../../test/pj-rehearse-integration/expected_jobs.yaml"
+)
+
+var (
```
stevekuznetsov: Why are these globals?

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,130 @@
...
+	expectedJobs := getExpectedProwJobs(t)
+	fakecs.Fake.PrependWatchReactor("prowjobs", func(clienttesting.Action) (bool, watch.Interface, error) {
+		watcher.Stop()
+		ret := watch.NewFakeWithChanSize(expectedJobs.Len(), true)
+
+		for event := range watcher.ResultChan() {
+			pj := event.Object.(*v1.ProwJob).DeepCopy()
+
+			if !expectedJobs.Has(pj.Spec.Job) {
+				t.Fatalf("Job %s shouldn't be rehearsed", pj.Spec.Job)
```
stevekuznetsov, line 80: Do we test the obverse? That all jobs we want to see rehearsed, actually get rehearsed?

stevekuznetsov (commented):
Two questions

cmd/pj-rehearse/main_test.go
droslean: addressed

cmd/pj-rehearse/main_test.go
droslean: addressed

petr-muller:
/lgtm
/hold

My concerns are resolved, holding for @stevekuznetsov

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,149 @@
...
+			}
+
+			if rehearsed.Has(pj.Spec.Job) {
+				t.Fatalf("Job %s has already been rehearsed", pj.Spec.Job)
+			}
+
+			pj.Status.State = v1.SuccessState
+			ret.Modify(pj)
+
+			toBeRehearsed.Delete(pj.Spec.Job)
```
stevekuznetsov: So technically you don't need to do this since you're checking `rehearsed.Has`.

cmd/pj-rehearse/main_test.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+		return true, ret, nil
+	})
+
+	executor := rehearse.NewExecutor(changedPresubmits, jobSpec.Refs.Pulls[0].Number, candidatePath, jobSpec.Refs, false, loggers, fakeclient)
+	if _, err := executor.ExecuteJobs(); err != nil {
+		t.Fatalf("Failed to execute rehearsal jobs: %v", err)
+	}
+
+	if toBeRehearsed.Len() != 0 {
```
stevekuznetsov: Instead of removing stuff from the expected list, just use set operations:

```go
if missing := rehearsed.Difference(toBeRehearsed); missing.Len() > 0 {
	t.Errorf("%s: did not rehearse expected prowJobs: %v", tc.name, missing.List())
}
if extra := toBeRehearsed.Difference(rehearsed); extra.Len() > 0 {
	t.Errorf("%s: found unexpectedly rehearsed prowJobs: %v", tc.name, extra.List())
}
```

stevekuznetsov (commented):
nit on set operations otherwise LGTM

cmd/pj-rehearse/main_test.go
droslean: addressed

cmd/pj-rehearse/main_test.go
droslean: addressed

PR #68, opened 2019-02-04
Improve failure mode messaging to the user

petr-muller:
Rehearse tool has 4 possible results:

1. There is nothing that we can rehearse. We can consider this case to be a success.
2. We attempted to rehearse some jobs and they all pass. This is also a success.
3. We attempted to rehearse some jobs and some of them failed. The whole purpose of rehearsal is to communicate these failures to the user.
4. We attempted to rehearse some jobs but failed to do so for some reason. This is the tricky case - failing on these will frustrate the users because it's usually not *their* error, but we cannot simply eat these errors because we would fake passing tests while the tests were not even executed correctly.

This change triages these failure modes and accompanies them with an
explaining outputs.

/cc @droslean @bbguimaraes @stevekuznetsov

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -70,7 +70,25 @@ func validateOptions(o options) error {
 	return nil
 }
 
-func gracefulExit(suppressFailures bool) {
+const (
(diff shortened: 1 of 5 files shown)


stevekuznetsov (approved):
/lgtm

PR #69, opened 2019-02-05
Properly propagate more failures and unit test all the things!

petr-muller:
Previously, when we failed to submit some rehearsal jobs, that failure could be covered up later by all successfully submitted rehearsals passing.

Additionally, added a bunch of unit tests for `ExecuteJobs` to do more testing of various, mostly negative, cases.

/cc @droslean @stevekuznetsov @bbguimaraes

PR #71, opened 2019-02-06
Embed branch in rehearsed job's context

petr-muller:
Previously, when a PR was attempting to add two jobs with same context,
but different branch, pj-rehearse would generate two rehearsal jobs with
the identical context, leading to both of them reporting into the same
GH context. The fix is to include target branch name into a context to
distinguish them.

Builds over https://github.com/openshift/ci-operator-prowgen/pull/69

### DIFF
--- a/pkg/rehearse/jobs_test.go
+++ b/pkg/rehearse/jobs_test.go
@@ -66,7 +66,6 @@ func makeCMReference(cmName, key string) *v1.EnvVarSource {
 
 func TestInlineCiopConfig(t *testing.T) {
 	testTargetRepo := "org/repo"
-	testLoggers := Loggers{logrus.New(), logrus.New()}
 
 	testCases := []struct {
 		description   string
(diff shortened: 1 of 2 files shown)


pkg/rehearse/jobs_test.go
```diff
@@ -144,33 +144,43 @@ func makeTestingPresubmit(name, context string, ciopArgs []string) *prowconfig.P
...
 	}{{
-		source: makeTestingPresubmit("pull-ci-openshift-ci-operator-master-build", "ci/prow/build", []string{"arg", "arg"}),
-		pr:     123,
+		description: "sanity test",
+		source: makeTestingPresubmit(
+			"pull-ci-openshift-ci-operator-master-build",
+			"ci/prow/build",
+			[]string{"arg", "arg"},
+			"master"),
 		expected: makeTestingPresubmit(
```
stevekuznetsov, line 167: Using helpers like `makeTestingPresubmit` make the tests more dependent on each other and fragile. I think unless we have a good reason we should use literals where possible. This is not something you need to tackle in this PR

stevekuznetsov (commented):
/lgtm
/hold

pkg/rehearse/jobs_test.go
petr-muller: I already started rewriting that once but did not finish because I got bogged down in changes. 

I'd like to avoid crafting `Presubmit` literals everywhere - they are quite deep structures, it's easy to make mistakes because stuff needs to match in them, and most tests really do not care about most fields in them, they just need "sane" structs.

At the same time, what you said is valid and the current state is certainly not ideal.

PR #72, opened 2019-02-07
Generate trigger regex as upstream does

stevekuznetsov:
We should not be matching the literal `/test all` with our triggers.
It's not best practice and should be avoided. This patch refactors our
trigger generation to use the same logic as the upstream defaulting
would, generating the identical regex for a job name as would happen if
no regex was present in the Prow config at all. We still need to
generate our own as we want the name in the regex to be a subset of the
full name of the job.


stevekuznetsov:
```
$ time make install integration
go install ./cmd/...
test/integration/run.sh
[INFO] Generating Prow jobs...
[INFO] Validating generated Prow jobs...
[INFO] Determinizing Prow jobs...
[INFO] Validating determinized Prow jobs...
[INFO] Success!

real	0m0.446s
user	0m0.715s
sys	0m0.188s
```

Holy shit I love our test suite

PR #73, opened 2019-02-07
Move ExecuteJobs logic to an executor.

droslean:
Create an executor that holds all the logic of the execution of the jobs. 
Also using channels to gather the Statuses of the prowjobs in order to create a detailed report for each job.

WIP -> Fix tests.

@stevekuznetsov @petr-muller @bbguimaraes


cmd/pj-rehearse/main.go
```diff
@@ -161,17 +161,32 @@ func main() {
...
+
+	go executor.ExecuteJobs()
+
+	results := make(map[string]pjapi.ProwJobStatus)
+	for {
+		select {
+		case result := <-resultChan:
+			results[result.Name] = result.Status
+		case err := <-errorChan:
+			loggers.Job.WithError(err).Error("Error has occurred")
```
petr-muller: The failure is not propagated to the end result.

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+}
+
+// ExecuteJobs takes configs for a set of jobs which should be "rehearsed", and
+// creates the ProwJobs that perform the actual rehearsal. *Rehearsal* means
+// a "trial" execution of a Prow job configuration when the *job config* config
+// is changed, giving feedback to Prow config authors on how the changes of the
+// config would affect the "production" Prow jobs run on the actual target repos
+func (e *Executor) ExecuteJobs() {
+	pjs := e.submitRehearsals()
```
petr-muller: The failure to submit jobs is not propagated anywhere.

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+	refs       *pjapi.Refs
+	loggers    Loggers
+	pjclient   pj.ProwJobInterface
+	resultChan chan Result
+	errorChan  chan error
+	quit       chan struct{}
+}
+
+func NewExecutor(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int, prRepo string, refs *pjapi.Refs,
+	loggers Loggers, pjclient pj.ProwJobInterface, dryRun bool, resultChan chan Result, errorChan chan error, quit chan struct{}) *Executor {
```
petr-muller: 10 params API is a bit clunky, I would consider coalescing the channels into a struct, they are always used together anyway.

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
+package rehearse
```
petr-muller: I don't think this needs to live in a separate file, it's unnecessarily fragmented. It still handles jobs.

petr-muller (commented):
I am not 100% sold I want to do this, big part of that is that I am not sure I would like to do this *now* - I'm not comfortable doing anything that does not contribute to us actually releasing the feature.

Pros:
- The code structure is better. Really. It'll be easier to maintain.

Cons:
- It's not just a refactor, it changes the behavior, even removing some desirable features (inline)
- It does not get us closer to release - quite the opposite, because it'll be hard to finish remaining work with the rewrite in progress, plus it will take time to polish the refactor and rewrite all the tests basically from scratch.

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	if !e.dryRun && len(pjs) > 0 {
+		if err := e.waitForJobs(pjs, selector); err != nil {
+			e.errorChan <- err
+			return
+		}
+	}
+
+	e.quit <- struct{}{}
```
stevekuznetsov: Close your output channels and block on the receiver end for them to be closed, don't use a `quit` channel

pkg/rehearse/executor.go
stevekuznetsov: This is all package-local so you do not need an exported New method and you can just create a struct literal for the executor

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (e *Executor) waitForJobs(jobs sets.String, selector string) error {
+	for {
+		w, err := e.pjclient.Watch(metav1.ListOptions{LabelSelector: selector})
```
stevekuznetsov: Wasn't there list/watch with resourceversion code from @bbguimaraes somewhere?

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+			if !jobs.Has(pj.Name) {
+				continue
+			}
+
+			e.resultChan <- Result{
+				Name:   pj.Name,
+				Status: pj.Status,
+			}
+
+			switch pj.Status.State {
```
stevekuznetsov: This is consuming the results -- it should be reading them from the other side of the channel

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,147 @@
...
+		labels[k] = v
+	}
+
+	prowJob := pjutil.NewProwJob(pjutil.PresubmitSpec(*job, *e.refs), labels)
+	e.loggers.Job.WithFields(pjutil.ProwJobFields(&prowJob)).Info("Submitting a new prowjob.")
+
+	return e.pjclient.Create(&prowJob)
+}
+
+func (e *Executor) getExecutionSelector() (string, error) {
```
stevekuznetsov: Why is this useful as a method if it's only called once?

stevekuznetsov (commented):
Producer/consumer could be cleaned up a bit and channels closed appropriately, otherwise design seems reasonable. Agreed with @petr-muller  that this is the wrong set of work to be doing right now.

droslean:
I agree with @petr-muller that it doesn't get us closer to release. I will rework this PR, and not include the channel changes at all. I would prefer though to keep the executor and keep the same logic that we currently have to execute the jobs.

pkg/rehearse/executor.go
droslean: addressed

droslean:
With the last commit, I revert the logic to the current one without using channels. 
Before merging, I would like to squash and move the tests that correspond to the executor to its test file.

pkg/rehearse/executor.go
```diff
@@ -0,0 +1,140 @@
...
+			jobs.Delete(pj.Name)
+			if jobs.Len() == 0 {
+				return success, nil
+			}
+		}
+	}
+}
+
+func (e *Executor) submitRehearsals() (sets.String, error) {
+	var validationErrors []error
```
petr-muller: Why is this called `validationErrors`? This is doing no validation.

droslean: addressed

pkg/rehearse/jobs.go
```diff
@@ -168,89 +169,99 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
-		}
-		loggers.Job.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
-		pjs.Insert(created.Name)
+	pjs, err := e.submitRehearsals()
+	if err != nil {
+		submitSuccess = false
 	}
-	if !follow {
+
+	if !e.dryRun {
```
petr-muller: This change does not make sense, when `dryRun == true`, the function is supposed to return here, but it's the opposite. It was called `follow` for a reason, it described the behavior better.

pkg/rehearse/jobs.go
```diff
@@ -168,89 +169,99 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
-		pjs.Insert(created.Name)
+	follow := !e.dryRun
+
+	pjs, err := e.submitRehearsals()
+	if err != nil {
+		submitSuccess = false
 	}
-	if !follow {
+
+	if follow {
```
petr-muller: When `follow` is true, we *want* it to continue to the end of the function (to "follow" the jobs).

```suggestion
	if !follow {
```

pkg/rehearse/jobs.go
```diff
@@ -168,89 +169,99 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
-	for _, job := range rehearsals {
-		created, err := submitRehearsal(job, refs, loggers, pjclient)
-		if err != nil {
-			loggers.Job.WithError(err).Warn("Failed to execute a rehearsal presubmit")
-			submitSuccess = false
-			continue
-		}
-		loggers.Job.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
-		pjs.Insert(created.Name)
+	follow := e.dryRun
```
petr-muller: Again, we want to follow when we are **not** running in the dry mode.

cmd/pj-rehearse/main.go
```diff
@@ -161,8 +161,11 @@ func main() {
 			gracefulExit(o.noFail, "")
 		}
 	}
+
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
-	success, err := rehearse.ExecuteJobs(changedPresubmits, prNumber, o.candidatePath, jobSpec.Refs, !o.dryRun, loggers, pjclient)
+
+	executor := rehearse.NewExecutor(changedPresubmits, prNumber, o.candidatePath, jobSpec.Refs, !o.dryRun, loggers, pjclient)
```
petr-muller: Passing `!o.dryRun` (negated) does not make sense here, because it's received and saved as `dryRun`. It made sense before, because it was received as `follow`.

petr-muller:
This is getting slightly confusing, let me try to summarize what `follow` means in `ExecuteJobs`.

The point of `dryRun` is to not make modifying calls to external world (=not actually create any jobs). This is achieved on a client level, in `NewProwJobClient` where `Create` calls are backed up by a fake when the client was created with `dry==true`. All the remaining code does not care about `dry`, with the **ONE** exception of waiting on jobs to finish. With fake client the jobs never finish, so waiting on them never ends. So the original solution was to make `ExecuteJobs` to have a `follow` flag (which was set set to `!dryRun`), which when set to `false` made the method skip the waiting part.

PR #74, opened 2019-02-08
Add end-to-end test for pj-rehearse 

petr-muller:
Craft an `openshift/release`-like repo, as if prepared by cloneref.
Prepare a `JOB_SPEC` to instruct pj-rehearse to work over that repo (in
dry mode), and roughly validate its output (specific content of rehearsal
jobs is validated on unit test level).

In a separate commit, I have also adjusted integration test directories,
naming and `Makefile` targets to the fact that we now have multiple things
to test.

I have reused @droslean 's test data from https://github.com/openshift/ci-operator-prowgen/pull/67


petr-muller:
grr stupid old git

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,54 @@
...
+cp -R "${candidate_data}"/* .
+git add ci-operator cluster
+git commit -m "Candidate version of openshift/release" --quiet
+candidate_sha="$(git rev-parse HEAD)"
+popd >/dev/null
+
+echo "[INFO] Running pj-rehearse in dry-mode..."
+submitted_yamls="${workdir}/rehearsals.yaml"
+error_log="${workdir}/errors.log"
+export JOB_SPEC='{"type":"presubmit","job":"pull-ci-openshift-release-master-rehearse","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"release","base_ref":"master","base_sha":"'${base_sha}'","pulls":[{"number":1234,"author":"petr-muller","sha":"'${candidate_sha}'"}]}}'
```
stevekuznetsov: Oh boy

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,54 @@
...
+
+if ! pj-rehearse --dry-run=true --no-fail=false --candidate-path "${fake_openshift_release}" > "${submitted_yamls}" 2> "${error_log}"; then
+  echo "[ERROR] pj-rehearse failed:"
+  cat "${error_log}"
+  exit 1
+fi
+
+echo "[INFO] Validating created rehearsals"
+rehearsed_jobs="${workdir}/rehearsed_jobs"
+grep "^  job: rehearse-1234-pull-ci" "${submitted_yamls}" | cut -c '8-' | sort  > "${rehearsed_jobs}"
```
stevekuznetsov: While this is like 70% of what we want, could we invent a different type of dry-run mode or something that will either POST to a fake server we have or just spit out the YAML it would have POSTed so we can validate like 100% of the output here?

petr-muller: I considered it but decided not to because the lower-level details are tested in lower level tests. But I'm willing to negotiate :)

Dry-run is certainly worth fixing so it outputs a proper YAML (list of jobs instead of concatenated individual job YAMLs) so it can be parsed back - that's quite easy. We could then write a Python helper or something to validate it against some expected output, but we would need to deal with the dynamic fields so it's not ultra trivial, and I'm not sure it's worth it because we test the job->rehearsal translation on unit level.

test/pj-rehearse-integration/run.sh
petr-muller: Maybe I'll extract it to a function so it does not clutter the flow that much.

test/pj-rehearse-integration/run.sh
stevekuznetsov: Unit tests can't replace e2e! I think we should try -- dynamic fields can be elided with `diff --ignore-matching-lines`, we manage that with k8s resources often. Plus, if we're not actually talking to a server we won't be getting fields like `creationTimestamp` so we should have fewer dynamic fields.

petr-muller: Not replace but complement, which I think is the case here. But OK, I'll give it a shot.

petr-muller:
@stevekuznetsov PTAL now. I have iterated a bit on the shell script so it is slightly more legible. It now does a full diff over the output YAML, excluding some lines (but hell, that basic regex syntax...). Obtaining deterministic output also needed a change in how the YAML is actually produced by `pj-rehearse`.

pkg/rehearse/jobs.go
```diff
@@ -301,9 +308,11 @@ func (e *Executor) waitForJobs(jobs sets.String, selector string) (bool, error)
 	}
 }
 
-func (e *Executor) submitRehearsals() (sets.String, error) {
+func (e *Executor) submitRehearsals() (sets.String, []*pjapi.ProwJob, error) {
 	var errors []error
-	pjs := make(sets.String, len(e.rehearsals))
+	pj_names := make(sets.String, len(e.rehearsals))
```
stevekuznetsov: Prefer `pjNames` over `pj_names`

stevekuznetsov: I know you didn't change this, but `len(e.rehearsals)` is the worst case scenario for the size of this set, right? Why do we need to do something other than use the `sets.NewStrings()` constructor?

pkg/rehearse/jobs.go
```diff
@@ -230,19 +225,31 @@ func NewExecutor(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
...
+
 // ExecuteJobs takes configs for a set of jobs which should be "rehearsed", and
 // creates the ProwJobs that perform the actual rehearsal. *Rehearsal* means
 // a "trial" execution of a Prow job configuration when the *job config* config
 // is changed, giving feedback to Prow config authors on how the changes of the
 // config would affect the "production" Prow jobs run on the actual target repos
 func (e *Executor) ExecuteJobs() (bool, error) {
 	submitSuccess := true
-	pjs, err := e.submitRehearsals()
+	names, pjs, err := e.submitRehearsals()
```
stevekuznetsov: If you change to just return the prowjob structs, can you do the reduction to a set of strings here rather than in the `submitRehearsals()`?

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,64 @@
...
+readonly REHEARSED_JOBS="${WORKDIR}/rehearsals.yaml"
+echo "[INFO] Running pj-rehearse in dry-mode..."
+if ! pj-rehearse --dry-run=true --no-fail=false --candidate-path "${FAKE_OPENSHIFT_RELEASE}" > "${REHEARSED_JOBS}" 2> "${WORKDIR}/pj-rehearse-stderr.log"; then
+  echo "[ERROR] pj-rehearse failed:"
+  cat "${WORKDIR}/pj-rehearse-stderr.log"
+  exit 1
+fi
+
+echo "[INFO] Validating created rehearsals"
+if ! compare_to_expected "$REHEARSED_JOBS" > "${WORKDIR}/diff"; then
```
stevekuznetsov: FYI -- not an issue right now but the `if ! function_call` structure will remove your ability to influence control flow through the script via `set -o errexit` in the function call. We're just calling `diff` once so it's not an issue right now

stevekuznetsov: You can also capture the output without an intermediate file if you want:

```sh
if ! output="$( compare_to_expected "$REHEARSED_JOBS" )"; then
```

test/pj-rehearse-integration/expected_jobs.yaml
```diff
@@ -0,0 +1,398 @@
+- apiVersion: prow.k8s.io/v1
```
stevekuznetsov, line 1: How come `pj-rehearse` doesn't have the normal user-facing output?

test/pj-rehearse-integration/run.sh
stevekuznetsov: nit: `"${REHEARSED_JOBS}"`

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,64 @@
...
+  echo "[ERROR] pj-rehearse output differs from expected:"
+  cat "${WORKDIR}/diff"
```
stevekuznetsov: nit: prepending these with `[ERROR]` will get them highlighted in gubernator but if no other scripts in the repo are doing that then forget it

pkg/rehearse/jobs.go
```diff
@@ -44,12 +45,6 @@ func NewProwJobClient(clusterConfig *rest.Config, namespace string, dry bool) (p
...
 		pjcset := pjclientsetfake.NewSimpleClientset()
 		pjcset.Fake.PrependReactor("create", "prowjobs", func(action testing.Action) (bool, runtime.Object, error) {
-			pj := action.(testing.CreateAction).GetObject().(*pjapi.ProwJob)
-			jobAsYAML, err := yaml.Marshal(pj)
-			if err != nil {
-				return true, nil, fmt.Errorf("failed to marshal job to YAML: %v", err)
-			}
-			fmt.Printf("%s\n", jobAsYAML)
 			return false, nil, nil
 		})
```
bbguimaraes: The reactor doesn't serve any purpose anymore, it can be removed.

pkg/rehearse/jobs.go
bbguimaraes: It is the "worst case scenario" in the sense that it will not be entirely filled if errors occur when creating the `PJ`s, which is an exceptional case.

In all other cases the set will be exactly this size, hence the pre-allocation.

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# This test sets up a temporary testing repository
+# resembling `openshift/release` and tries to execute
+# pj-rehearse as if run over a PR for that repo
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+readonly WORKDIR="$( mktemp -d )"
```
bbguimaraes: _it looks like @stevekuznetsov forgot the -Wpedantic flag_

`readonly` will mask the result of `mktemp` in case of failure.

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,64 @@
...
+
+make_testing_repository() {
+  local -r master_data="${PWD}/${TEST_ROOT}/master"
+  local -r candidate_data="${PWD}/${TEST_ROOT}/candidate"
+
+  echo "[INFO] Preparing fake input repository..."
+  mkdir "${FAKE_OPENSHIFT_RELEASE}"
+  pushd "${FAKE_OPENSHIFT_RELEASE}" >/dev/null
+  git init --quiet
+  cp -R "${master_data}"/* .
```
bbguimaraes, line 27: I suspect `git clone`ing the original repository would be faster.

test/pj-rehearse-integration/run.sh
```diff
@@ -0,0 +1,64 @@
...
+  git add ci-operator cluster
+  git commit -m "Master version of openshift/release" --quiet
+  local -r base_sha="$(git rev-parse HEAD)"
```
bbguimaraes: Same as `readonly` above.

test/pj-rehearse-integration/expected_jobs.yaml
bbguimaraes: I was also wondering that. We could add a flag similar to `oc`'s `-o yaml` — which I imagine could be useful in other contexts ­— to avoid sacrificing the valuable default output, akin to what we did for `--debug-log`.

pkg/rehearse/jobs.go
petr-muller: Addressed.

pkg/rehearse/jobs.go
petr-muller: Addressed.

test/pj-rehearse-integration/run.sh
petr-muller: Addressed. I wanted to hide the diff params in the function. I really hope the script will not really grow beyond this.

test/pj-rehearse-integration/expected_jobs.yaml
petr-muller: It only has logging ATM... Maybe we could fix it but right now clear stdout is quite handy in dry-mode :)

test/pj-rehearse-integration/run.sh
petr-muller: Addressed.

pkg/rehearse/jobs.go
petr-muller: Addressed. Also gets rid of two imports which is nice.

test/pj-rehearse-integration/run.sh
petr-muller: Addressed. Apparently `readonly` works slightly differently than I expected. I actually rarely used it before. That's what you get for trying :/

test/pj-rehearse-integration/run.sh
petr-muller: I don't think we need to care about speed here, it's just a few files.

test/pj-rehearse-integration/run.sh
petr-muller: Addressed.

petr-muller:
/hold

Addressed the feedback in a separate commit for clarity. I'll squash before merge.

stevekuznetsov:
/lgtm
Squash away

petr-muller:
/hold cancel

I have no idea why the `lgtm` is gone, I just squashed last two commits :(

PR #75, opened 2019-02-08
Force rehearsals to have an unified trigger

petr-muller:
Rehearsals cannot be triggered directly (Prow will only kick jobs that
are configured for the given repo on /test XXX commands), so we
work around it by making it share RerunCommand with the `pj-rehearse`
job (the driver). RerunCommand is only used to give users hints in a
comment, so this just makes the bot stop suggesting comments that would
not work.

I've also simplified the affected test and made it less coupled with the others.

### DIFF
--- a/pkg/rehearse/jobs_test.go
+++ b/pkg/rehearse/jobs_test.go
@@ -7,6 +7,7 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/getlantern/deepcopy"
 	"github.com/sirupsen/logrus"
 	logrustest "github.com/sirupsen/logrus/hooks/test"
 
@@ -144,7 +145,8 @@ func makeTestingPresubmit(name, context string, ciopArgs []string, branch string
 				}},
 			},
 		},
-		Context: context,
+		RerunCommand: "/test pj-rehearse",
+		Context:      context,
 		Brancher: prowconfig.Brancher{Branches: []string{
 			fmt.Sprintf("^%s$", branch),
 		}},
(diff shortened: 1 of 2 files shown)


pkg/rehearse/jobs_test.go
```diff
@@ -2,6 +2,7 @@ package rehearse
 
 import (
 	"fmt"
+	"github.com/getlantern/deepcopy"
```
stevekuznetsov: nit: group imports by stdlib, then third-party, then project-local

pkg/rehearse/jobs_test.go
```diff
@@ -153,35 +155,40 @@ func makeTestingPresubmit(name, context string, ciopArgs []string, branch string
...
+		Context:      "ci/prow/test",
+		Brancher:     prowconfig.Brancher{Branches: []string{"^branch$"}},
 	}
+	expectedPresubmit := &prowconfig.Presubmit{}
+	deepcopy.Copy(expectedPresubmit, sourcePresubmit)
+
+	expectedPresubmit.Name = "rehearse-123-pull-ci-org-repo-branch-test"
+	expectedPresubmit.Labels = map[string]string{rehearseLabel: "123"}
+	expectedPresubmit.Spec.Containers[0].Args = []string{"arg1", "arg2", "--git-ref=org/repo@branch"}
+	expectedPresubmit.RerunCommand = "/test pj-rehearse"
```
stevekuznetsov, line 180: Prow enforces that the compiled trigger regex matches the rerun command. This may just be on config load, but test this e2e before we merge to make sure it won't break anything, or set the trigger appropriately.

pkg/rehearse/jobs_test.go
```diff
@@ -153,35 +155,40 @@ func makeTestingPresubmit(name, context string, ciopArgs []string, branch string
...
+	expectedPresubmit.Context = "ci/rehearse/org/repo/branch/test"
+
+	t.Run("sanity test", func(t *testing.T) {
```
stevekuznetsov: This is only useful for running tests in parallel, can remove

stevekuznetsov (commented):
Mostly LGTM

pkg/rehearse/jobs_test.go
petr-muller: I was aware of the check, and I tested it:

https://github.com/openshift/release/pull/2596#issuecomment-461896772

PR #76, opened 2019-02-11
unexport variables, also add some comments.

droslean:
Unexport constants that are currently being used only in this package. 
Also added some comments.


stevekuznetsov (approved):
/lgtm

PR #77, opened 2019-02-12
Strip CPU limits from prowjobs directly due to cfs_quota issues

smarterclayton:
cfs_quota has challenges when it comes to distributing CPU and we
currently suspect use of limits is causing CPU inversion (processes
stall for long periods of time without CPU). Remove the generated
limits from prowjobs.


petr-muller:
It'll need a buddy PR for `openshift/release` afterwards.

smarterclayton:
Yes

droslean:
Also, the integration data files have to change. `test/integration/data/`

smarterclayton:
Fixed tests

PR #78, opened 2019-02-13
Limit rehearsals to 15 jobs

petr-muller:
I'd like to consider something like this to go along fully enabling rehearsals in https://github.com/openshift/release/pull/2847

I'm concerned whether it's fine to mass-submit hundreds of prowjobs for changes like yesterday's https://github.com/openshift/release/pull/2834. It seems that GH has a limit of 1000 statuses per SHA so we're fine on that front, but I'm not sure about our test throughput. Also, we'll probably not care that much about jobs themselves when we do mass changes.

WDYT?

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -45,10 +45,8 @@ type options struct {
 	noFail       bool
 	debugLogPath string
 
-	configPath    string
-	jobConfigPath string
-
-	candidatePath string
+	candidatePath  string
+	rehearsalLimit int
 }
 
 func gatherOptions() options {
@@ -60,6 +58,7 @@ func gatherOptions() options {
 	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
 
 	fs.StringVar(&o.candidatePath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
+	fs.IntVar(&o.rehearsalLimit, "rehearsal-limit", 15, "Upper limit of jobs attempted to rehearse (if more jobs would be rehearsed, none will)")
 
 	fs.Parse(os.Args[1:])
 	return o
@@ -165,6 +164,16 @@ func main() {
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
 	executor := rehearse.NewExecutor(changedPresubmits, prNumber, o.candidatePath, jobSpec.Refs, o.dryRun, loggers, pjclient)
+
+	if executor.RehearsalJobCount > o.rehearsalLimit {
+		jobCountFields := logrus.Fields{
+			"rehearsal-threshold": o.rehearsalLimit,
+			"rehearsal-jobs":      executor.RehearsalJobCount,
+		}
+		logger.WithFields(jobCountFields).Info("Would rehearse too many jobs, will not proceed")
+		os.Exit(0)
+	}
+
 	success, err := executor.ExecuteJobs()
 	if err != nil {
 		logger.WithError(err).Error("Failed to rehearse jobs")
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -206,6 +206,8 @@ func configureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prR
 
 // Executor holds all the information needed for the jobs to be executed.
 type Executor struct {
+	RehearsalJobCount int
+
 	dryRun     bool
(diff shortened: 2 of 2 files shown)


pkg/rehearse/jobs.go
```diff
@@ -220,6 +222,8 @@ func NewExecutor(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
 	dryRun bool, loggers Loggers, pjclient pj.ProwJobInterface) *Executor {
 	rehearsals := configureRehearsalJobs(toBeRehearsed, prRepo, prNumber, loggers)
 	return &Executor{
+		RehearsalJobCount: len(rehearsals),
+
```
droslean, line 226: nit: no need for the extra line here

droslean (approved):
I like this. But we should agree on the threshold limit.  15 sounds ok to me.

pkg/rehearse/jobs.go
petr-muller: I wanted to separate private and exported fields

petr-muller:
Yeah, 15 is entirely arbitrary :)

bbguimaraes:
+0, we only rehearse container tests for now, so the impact shouldn't be that large, but it's nice to have a safety net.

droslean:
lgtm from my side. @stevekuznetsov wanna check?

cmd/pj-rehearse/main.go
```diff
@@ -86,6 +86,7 @@ an infrastructure issue.`
 
 pj-rehearse rehearsed jobs and at least one of them failed. This means that
 job would fail when executed against the current HEAD of the target branch.`
+	REHEARSAL_THRESHOLD = 15
```
stevekuznetsov: Go prefers `rehearsalThreshold`

cmd/pj-rehearse/main.go
```diff
@@ -165,6 +166,16 @@ func main() {
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
 	executor := rehearse.NewExecutor(changedPresubmits, prNumber, o.candidatePath, jobSpec.Refs, o.dryRun, loggers, pjclient)
+
+	if executor.RehearsalJobCount > REHEARSAL_THRESHOLD {
```
stevekuznetsov: Are we sorting the jobs? For the same set of selected jobs for rehearsal, will we choose the same 15 every time?

stevekuznetsov (commented):
Would it be hard to add a `--limit` flag to this so we can tweak  it after the fact by changing just job config

cmd/pj-rehearse/main.go
petr-muller: > +0, we only rehearse container tests for now, so the impact shouldn't be that large, but it's nice to have a safety net.

Yeah, it's not that we would attempt to build hundreds of clusters, but [Clayton's PR yesterday](https://github.com/openshift/release/pull/2834) would submit ~320 jobs at once.

petr-muller:
> +0, we only rehearse container tests for now, so the impact shouldn't be that large, but it's nice to have a safety net.

Yeah, it's not that we would attempt to build hundreds of clusters, but [Clayton's PR yesterday](https://github.com/openshift/release/pull/2834) would still submit ~320 jobs (container-based but still) at once, which sounds excessive.

cmd/pj-rehearse/main.go
petr-muller: I'm not attempting to sample, I'm bailing out when too many jobs would be rehearsed.

cmd/pj-rehearse/main.go
petr-muller: Aaaargh capitalized constants are burned to my brain

petr-muller:
I'll make it a parameter with a default.

petr-muller:
@stevekuznetsov PTAL now

pkg/rehearse/jobs.go
```diff
@@ -220,6 +222,8 @@ func NewExecutor(toBeRehearsed map[string][]prowconfig.Presubmit, prNumber int,
 	dryRun bool, loggers Loggers, pjclient pj.ProwJobInterface) *Executor {
 	rehearsals := configureRehearsalJobs(toBeRehearsed, prRepo, prNumber, loggers)
```
droslean, line 223: Since we add more stuff to the constructor, I am starting thinking that it's better to export this function and call it from `main` and pass the rehearsals to the constructor directly. Then we don't have to include the `RehearsalJobCount` in the executor and we can make the limit check in main without creating the `executor`.

petr-muller: I think when we have an extra layer already, let's use it to hide some details.

PR #80, opened 2019-02-15
add optional:true to rehearsal jobs

droslean:


petr-muller:
Post-factum note: This does not affect anything. Tide only considers jobs configured for a repository when it computes it's optional/required contexts sets, and rehearsals are basically treated as non-prow contexts. https://github.com/openshift/release/pull/2864 is one option how to make rehearsals really optional.

PR #81, opened 2019-02-15
Add `allow-volumes` flag. Split filtering logic from the executor

droslean:
`--allow-volumes` flag added. 
Splits the filtering logic from the executor.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -43,6 +43,7 @@ func loadClusterConfig() (*rest.Config, error) {
 type options struct {
 	dryRun       bool
 	noFail       bool
+	allowVolumes bool
 	debugLogPath string
 
 	candidatePath  string
@@ -55,9 +56,11 @@ func gatherOptions() options {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 	fs.BoolVar(&o.noFail, "no-fail", true, "Whether to actually end unsuccessfuly when something breaks")
-	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
+	fs.BoolVar(&o.allowVolumes, "allow-volumes", false, "Allows jobs with extra volumes to be rehearsed")
 
+	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
 	fs.StringVar(&o.candidatePath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
+
 	fs.IntVar(&o.rehearsalLimit, "rehearsal-limit", 15, "Upper limit of jobs attempted to rehearse (if more jobs would be rehearsed, none will)")
 
 	fs.Parse(os.Args[1:])
@@ -147,12 +150,6 @@ func main() {
 		gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
-	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
-	if len(changedPresubmits) == 0 {
-		logger.Info("no jobs to rehearse have been found")
-		os.Exit(0)
-	}
-
 	debugLogger := logrus.New()
 	debugLogger.Level = logrus.DebugLevel
 	if o.debugLogPath != "" {
@@ -164,20 +161,23 @@ func main() {
 			gracefulExit(o.noFail, "")
 		}
 	}
-
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
-	executor := rehearse.NewExecutor(changedPresubmits, prNumber, o.candidatePath, jobSpec.Refs, o.dryRun, loggers, pjclient)
-
-	if executor.RehearsalJobCount > o.rehearsalLimit {
+	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
+	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, o.candidatePath, prNumber, loggers, o.allowVolumes)
+	if len(rehearsals) == 0 {
+		logger.Info("no jobs to rehearse have been found")
+		os.Exit(0)
+	} else if len(rehearsals) > o.rehearsalLimit {
 		jobCountFields := logrus.Fields{
 			"rehearsal-threshold": o.rehearsalLimit,
-			"rehearsal-jobs":      executor.RehearsalJobCount,
+			"rehearsal-jobs":      len(rehearsals),
 		}
 		logger.WithFields(jobCountFields).Info("Would rehearse too many jobs, will not proceed")
 		os.Exit(0)
 	}
 
+	executor := rehearse.NewExecutor(rehearsals, prNumber, o.candidatePath, jobSpec.Refs, o.dryRun, loggers, pjclient)
 	success, err := executor.ExecuteJobs()
 	if err != nil {
 		logger.WithError(err).Error("Failed to rehearse jobs")
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -56,10 +56,6 @@ func NewProwJobClient(clusterConfig *rest.Config, namespace string, dry bool) (p
 }
 
 func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber int) (*prowconfig.Presubmit, error) {
-	if err := filterJob(source); err != nil {
-		return nil, err
-	}
-
 	var rehearsal prowconfig.Presubmit
 	deepcopy.Copy(&rehearsal, source)
 
@@ -82,7 +78,22 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber 
 	return &rehearsal, nil
 }
 
-func filterJob(source *prowconfig.Presubmit) error {
+func filterJobs(changedPresubmits map[string][]prowconfig.Presubmit, allowVolumes bool, logger logrus.FieldLogger) map[string][]prowconfig.Presubmit {
+	ret := make(map[string][]prowconfig.Presubmit)
+	for repo, jobs := range changedPresubmits {
+		for _, job := range jobs {
+			jobLogger := logger.WithFields(logrus.Fields{"repo": repo, "job": job.Name})
+			if err := filterJob(&job, allowVolumes); err != nil {
+				jobLogger.WithError(err).Warn("could not rehearse job")
+				continue
+			}
+			ret[repo] = append(ret[repo], job)
+		}
+	}
+	return ret
+}
+
+func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
 	// there will always be exactly one container.
 	container := source.Spec.Containers[0]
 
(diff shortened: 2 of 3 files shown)


pkg/diffs/diffs_test.go
```diff
@@ -148,3 +147,59 @@ func makeConfig(p []prowconfig.Presubmit) *prowconfig.Config {
...
+		crippleFunc: func(j *prowconfig.Presubmit) {
+			j.Spec.Containers[0].Args = append(j.Spec.Containers[0].Args, "--git-ref=organization/repo@master")
+		},
+	}, {
+		description: "jobs running over multiple branches",
+		crippleFunc: func(j *prowconfig.Presubmit) {
+			j.Brancher.Branches = append(j.Brancher.Branches, "^feature-branch$")
+		},
+	}, {
+		description: "jobs that need additional volumes mounted",
```
stevekuznetsov: Let's test the other case, too

cmd/pj-rehearse/main.go
```diff
@@ -147,7 +150,11 @@ func main() {
 		gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
-	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
+	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger, o.allowVolumes)
+	if len(changedPresubmits) == 0 {
+		logger.Info("No jobs have been found to rehearse")
```
stevekuznetsov: Why?

droslean: if there are no jobs found to rehearse, there is no reason to continue, right?

stevekuznetsov: But we're handling it already, and we even have tests to that effect. Why change it?

stevekuznetsov:
@droslean FYI -- not a criticism but this is another PR where you're proposing more than one change. If you have to use "and" or multiple sentences to describe what a PR does, try breaking it up into smaller ones that are focused -- we can merge those quicker :)

droslean:
@stevekuznetsov I think we discuss adding this flag in our meeting, in order to start working on the templates. The extra changes are to pass the flag correctly to the diffs instead of the executor.  But you are right! I should break this PR to smaller ones. Sorry for the inconvenience.

stevekuznetsov:
The executor diff makes sense, I meant these two ;)

 - Also added a check to exit if there are no jobs to rehearse.
 - nits: group all const together.

pkg/diffs/diffs.go
```diff
@@ -41,13 +41,23 @@ func GetChangedPresubmits(prowMasterConfig, prowPRConfig *prowconfig.Config, log
 			if job.Agent == string(pjapi.KubernetesAgent) {
 				// If the agent was changed and is a kubernetes agent, just choose the job for rehearse.
 				if masterJob.Agent != job.Agent {
+					if err := filterJob(&job, allowVolumes); err != nil {
+						logger.WithFields(logFields).WithError(err).Warn("Skiping job")
```
petr-muller: Nit: `Skiping` ⇨ `Skipping`

pkg/diffs/diffs.go
```diff
@@ -41,13 +41,23 @@ func GetChangedPresubmits(prowMasterConfig, prowPRConfig *prowconfig.Config, log
...
 					ret[repo] = append(ret[repo], job)
 					continue
 				}
 
 				if !equality.Semantic.DeepEqual(masterJob.Spec, job.Spec) {
+					if err := filterJob(&job, allowVolumes); err != nil {
+						logger.WithFields(logFields).WithError(err).Warn("Skiping job")
+						continue
+					}
+
```
petr-muller: Adding a new if block highlighted the fact that we should deduplicate the code so we do not need to change two places anymore.

Just saying, it does not need to be addressed in this PR.

pkg/diffs/diffs_test.go
```diff
@@ -85,15 +86,15 @@ func TestGetChangedPresubmits(t *testing.T) {
 			configGenerator: func() (*prowconfig.Config, *prowconfig.Config) {
 				var p []prowconfig.Presubmit
 				deepcopy.Copy(&p, basePresubmit)
-				p[0].Spec.Containers[0].Command = []string{"test-command"}
+				p[0].Spec.Containers[0].Args = []string{"test-arg"}
```
petr-muller: Why are these changes needed?

pkg/diffs/diffs_test.go
```diff
@@ -148,3 +147,82 @@ func makeConfig(p []prowconfig.Presubmit) *prowconfig.Config {
...
+				Containers: []v1.Container{{
+					Command: []string{"ci-operator"},
+					Args:    []string{"arg"},
+				}},
+			},
+		},
+		RerunCommand: "/test pj-rehearse",
+		Context:      "ci/prow/test",
+		Brancher: prowconfig.Brancher{Branches: []string{
+			fmt.Sprintf("^%s$", "master"),
```
petr-muller: Why the `Sprintf` here, can be straight `"^master$"`

petr-muller (commented):
I have few inline comments and mostly minor code feedback. But the main thing for me here is I don't understand **why** the filtering-out logic needed to be moved from `jobs.makeRehearsalPresubmit` to `diffs.GetChangedPresubmit`, and without a good reason, I am currently not convinced it is a proper thing to do. The same functionality can be achieved by propagating the flag boolean into a different place.

Reasons why I like the current structure and dislike the new one:

1. `makeRehearsalPresubmit` is coupled to the filtering. It can only make rehearsals for *some* presubmits. Previously, it was able to handle any input presubmit by doing the filtering out itself, and erroring out on those for which a correct rehearsal could not be produced. To me, this is a good contract. After a change, `makeRehearsalPresubmit` relies blindly that whatever is passed to it is a "rehearsable" job, and what is "rehearsable" is determined by a totally unrelated method called somewhere distant in the code.
2. `GetChangedPresubmits` name suggests that it does exactly that, but after the change, that name no longer holds. It could be changed to `GetRehearsablePresubmits` or something, but I would prefer to keep the method's responsibility the way it is.

droslean:
@petr-muller Thanks for the review.  My intention by moving the filtering to `diffs`, was to avoid filtering each job inside the `executor`.  IMHO that logic should not be in the executor.  Perhaps `GetChangedPresubmit` should be renamed also. 

I would prefer to keep the `executor` simple. To just execute the jobs that we pass to it.

@stevekuznetsov @bbguimaraes any thought? I can revert the logic back.

petr-muller:
@droslean I think I understand the problem you have with the executor. I can be convinced it is currently doing "too much" and I can imagine us reaping the non-executing parts out of it. But as proposed, this change introduces coupling between seemingly unrelated code, which to me is a serious problem - I think the filtering and `makeRehearsalPresubmit` need to stay close.

We could consider reaping the whole `configureRehearsalPresubmits` out of executor, and we have two options here. In one, our main would be structured like this:
```
changedJobs := GetChangedPresubmits(old, new)
rehearsals := GetRehearsals(changedJobs)
execute(rehearsals)
```

The other could look like:
```
rehearsals := GetRehearsals(old, new)
execute(rehearsals)
```

The former is quite straightforward. In the latter, `GetRehearsals` would need to be internally structured well to properly separate the "detect changes" and "make rehearsals" concerns.

stevekuznetsov:
Especially as our filtering code gets complex I think I agree with both @droslean and @petr-muller -- we should have an explicit part of our pipeline for filtering and it should live as an intermediate between the change detector and the executor. We should think about a nice structure to allow us to extend filters in the future, as we add more.

petr-muller (commented):
Unfortunately my main concerns about distant code coupling were not addressed, and because the changes are not accompanied by any description, I do not know whether it's intentional or not. It's hard to discuss it that way.

Thinking about just "filtering" is just half of the picture - the filtering is only there to weed out jobs for which `makeRehearsalPresubmit` cannot produce anything meaningful, no other reason. They are coupled. I don't think we should have a top-level `Filter` API, I think we should have `MakeRehearsals` API (or something like that).

droslean:
After a meeting we have with @stevekuznetsov @petr-muller @bbguimaraes, we decide to split the logic of filtering the jobs from the executor. This PR was changed to support this implementation.

pkg/rehearse/jobs_test.go
```diff
@@ -640,7 +598,7 @@ func TestWaitForJobs(t *testing.T) {
 				return true, w, nil
 			})
 
-			executor := NewExecutor(nil, 0, "", &pjapi.Refs{}, true, loggers, cs.ProwV1().ProwJobs("test"))
+			executor := NewExecutor(nil, 0, "", &pjapi.Refs{}, false, loggers, cs.ProwV1().ProwJobs("test"))
```
stevekuznetsov: What's the meaning of this?

pkg/rehearse/jobs.go
```diff
@@ -163,11 +174,18 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
 	return &rehearsal, nil
 }
 
-func configureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prRepo string, prNumber int, loggers Loggers) []*prowconfig.Presubmit {
+// ConfigureRehearsalJobs filters the jobs that should be rehearsed, then return a list of them re-configured with the
+// ci-operator's configuration inlined.
+func ConfigureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, prRepo string, prNumber int, loggers Loggers, allowVolumes bool) []*prowconfig.Presubmit {
+	rehearsalsFiltered := filterJobs(toBeRehearsed, allowVolumes, loggers.Job)
+	if len(rehearsalsFiltered) == 0 {
+		return nil
```
stevekuznetsov: Why? Doesn't `filterJobs` return `nil`?

pkg/rehearse/jobs_test.go
droslean: Oops, I didn't again :). I include these changes in #87 . I will remove them from this PR.

pkg/rehearse/jobs.go
droslean: While filtering, some jobs can be excluded. Here, we make sure that if we exclude all the jobs, not to continue.

stevekuznetsov: Not a fan of this or the other change you have made to explicitly exit early on empty slices. `for` loop over a slice will do nothing if there is nothing in the slice. No need to have this be explicit.

droslean: Yes, good point. I will change it.

pkg/rehearse/jobs_test.go
```diff
@@ -347,7 +302,8 @@ func TestExecuteJobsErrors(t *testing.T) {
 				return false, nil, nil
 			})
 
-			executor := NewExecutor(tc.jobs, testPrNumber, testRepoPath, testRefs, true, testLoggers, fakeclient)
+			rehearsals := ConfigureRehearsalJobs(tc.jobs, testRepoPath, testPrNumber, testLoggers, true)
+			executor := NewExecutor(rehearsals, testPrNumber, testRepoPath, testRefs, false, testLoggers, fakeclient)
```
petr-muller: I'm not a fan of how these tests are actually silently changed (`dryRun` param). I know what you want to achieve here and the change actually improves the coverage, but it would deserve a separate commit with explanation, not to be buried with a refactor/unrelated feature commit.

petr-muller (approved):
I think we can merge this (leaving it to @stevekuznetsov who had some comments to be addressed), but given it actually did three distinct changes (allow-volumes feature, split the filtering logic, changed unrelated tests) I'd really prefer the changes in at least separate commits, and maybe even in separate PRs - it's mostly a feedback for next time.

pkg/rehearse/jobs_test.go
droslean: I agree with your point. I will exclude the `dryRun` param change from this PR. This change should be covered in #87

PR #82, opened 2019-02-17
Images in the 'ocp' namespace should have the release payload

smarterclayton:
Remove the 'artifacts: images' label, it has been replaced by other
constructs. Clarify what is legacy and what is not with regards to
namespace rules.

We are removing 'openshift/origin-v4.0' in favor of 'ocp/4.0' and
'ocp/4.X' in the future.

Prereq for openshift/release#2878

### DIFF
--- a/GENERATOR.md
+++ b/GENERATOR.md
@@ -65,19 +65,6 @@ promote the component images built by this postsubmit.
     ...
 ```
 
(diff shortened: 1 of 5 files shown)


stevekuznetsov (approved):
/lgtm

PR #87, opened 2019-02-20
[WIP] improve rehearse tests.

droslean:
The rehearse tests were running in dry-mode and didn't make use of the corresponding fake client.

### DIFF
--- a/pkg/rehearse/jobs_test.go
+++ b/pkg/rehearse/jobs_test.go
@@ -305,7 +305,7 @@ func makeSuccessfulFinishReactor(watcher watch.Interface, jobs map[string][]prow
 	}
 }
 
-func TestExecuteJobsErrors(t *testing.T) {
+func TestSubmitRehearsalsErrors(t *testing.T) {
 	testPrNumber, testNamespace, testRepoPath, testRefs := makeTestData()
 	targetRepo := "targetOrg/targetRepo"
 
@@ -333,11 +333,7 @@ func TestExecuteJobsErrors(t *testing.T) {
 			testLoggers := Loggers{logrus.New(), logrus.New()}
 			fakecs := fake.NewSimpleClientset()
 			fakeclient := fakecs.ProwV1().ProwJobs(testNamespace)
-			watcher, err := fakeclient.Watch(metav1.ListOptions{})
-			if err != nil {
-				t.Fatalf("Failed to setup watch: %v", err)
-			}
-			fakecs.Fake.PrependWatchReactor("prowjobs", makeSuccessfulFinishReactor(watcher, tc.jobs))
+
 			fakecs.Fake.PrependReactor("create", "prowjobs", func(action clientgo_testing.Action) (bool, runtime.Object, error) {
 				createAction := action.(clientgo_testing.CreateAction)
 				pj := createAction.GetObject().(*pjapi.ProwJob)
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Shouldn't there have been failing tests? Can we write a test case that fails without this patch and succeeds after it?

droslean:
Those tests were created for dry-run false in the first place.  It seems that we forgot them when we change the way we pass the `dry-run` to the executor.

stevekuznetsov:
My point stands -- if this is important to the tests, and they did not fail, we have bad tests.

droslean:
Yes exactly! I will do a follow-up PR to fix them.

bbguimaraes:
I agree with @stevekuznetsov's first comment. Before we make any changes to the code, I think it is important to understand why no test failed if this is an incorrect configuration.

droslean:
I will check, and include the changes in this PR

droslean:
@stevekuznetsov @petr-muller @bbguimaraes 
The tests are correct. We have also negative tests, but they didn't fail because of [this](https://github.com/openshift/ci-operator-prowgen/blob/master/pkg/rehearse/jobs.go#L251). So, it turns out that the tests are not testing the whole functionality of `executeJobs`. We have to decide how the executor will manage the pjclient in dry-mode or the opposite.

droslean:
The negative test that we have, is running already in `dry-run=false`. Thats why we didn't notice the flake in the positive test.

petr-muller:
I think this shows us that renaming the original `follow` arg of  `ExecuteJobs` to `dryRun` was a mistake and it's now confusing. So I propose to remove the `dryRun` member of the executor struct and make `follow` an `ExecuteJobs` argument again (that method is the only place where current `dryRun` is used). That way its semantics becomes clear again, and only tests that deal with `ExecuteJobs`  will need to work with it. 

I do not agree with repurposing `TestExecuteJobsErrors` into a test for `TestSubmitRehearsals` - having a test for `submitRehearsal` is a good thing but we still need a test for error conditions in `ExecuteJobs`. 

If we want to have a complete-ish coverage of `ExecuteJobs`, we'll need these cases:

```
1  follow=true  | all jobs fail to submit -> success=true, error!=nil
2  follow=false | all jobs fail to submit -> success=true, error!=nil
3  follow=true  | some jobs fail to submit | all submitted jobs succeed -> success=true, error!=nil
4  follow=false | some jobs fail to submit | all submitted jobs succeed -> success=true, error!=nil
5  follow=true  | some jobs fail to submit | some submitted jobs fail -> success=false, error!=nil 
6  follow=false | some jobs fail to submit | some submitted jobs fail -> success=true, error!=nil
7  follow=true  | all jobs submit | some submitted jobs fail -> success=false, error=nil
8  follow=false | all jobs submit | some submitted jobs fail -> success=true, error=nil
9  follow=true  | all jobs submit | all submitted jobs succeed -> success=true, error=nil
10 follow=false | all jobs submit | all submitted jobs succeed -> success=true, error=nil
```
And in most cases, we'll want to validate what jobs were actually submitted.

Few more notes:
- We probably need `follow=true` tests a bit more than `follow=false` ones
- Currently, we test cases 2, 4, 7, 10 (note how this contradicts the previous point)
- In some cases `follow` by itself does not change the expected results (that's why flipping it in `TestExecuteJobsPositive` did not affect the result, but in fact improved the test - it turned case 10 into case 9)
- `TestExecuteJobsPositive` is much more a test of `configureRehearsalJobs` than it is for `ExecuteJobs`.

So if we want to improve our tests, I think we have the following actions we may do, preferably in a series of smaller targeted PRs:

1. Finish extraction of non-executing functionality (`configureRehearsalJobs`) from the executor - this should include testcases from `TestExecuteJobsPositive`, because the are really focused on that functionality.
2. Reintroduce `follow` into `ExecuteJobs`
3. Fill the gaps in coverage, if desired
4. Write separate tests for `submitRehearsals`

droslean:
I agree with @petr-muller in some points only.  We already have the `follow` logic introduced in the `executor`, with the `dry-run`.  Also IMHO, we don't need to have tests that are not `following`, since our `dry-run` just only shows the prowjobs as yaml. All tests should be treated as `dry-run=false` and make good use of fake clients.
 I agree that we have to improve our tests and add more cases.

petr-muller:
> We already have the `follow` logic introduced in the `executor`, with the `dry-run`.

The problem is that we have `follow` logic, achieved via `dryRun`-called boolean. At the same time, the actual dry-run (not make modifying calls to external systems) logic is implemented elsewhere (in a construction of a client, which is simply passed to `ExecuteJobs`), and the `follow` logic is the only thing that `dryRun` boolean affects. That is wrong.

> Also IMHO, we don't need to have tests that are not `following`, since our `dry-run` just only shows the prowjobs as yaml. All tests should be treated as `dry-run=false` and make good use of fake clients.

Given the `follow` significantly affects what `ExecuteJobs` does, we should have at least some tests exercising that behavior.

droslean:
> The problem is that we have `follow` logic, achieved via `dryRun`-called boolean. At the same time, the actual dry-run (not make modifying calls to external systems) logic is implemented elsewhere (in a construction of a client, which is simply passed to `ExecuteJobs`), and the `follow` logic is the only thing that `dryRun` boolean affects. That is wrong.

Agreed. Shall we construct the `pjclient` inside the `executor`?

droslean:
I am closing this PR, since we are still making changes and adding templates and ci-operator configs so the tests will be in progress.

PR #88, opened 2019-02-21
 Add a utility to branch configurations for new releases 

stevekuznetsov:

---

/assign @smarterclayton 
/cc @petr-muller @droslean @bbguimaraes


stevekuznetsov:
(this is going to fail gofmt since I'm on 1.11 ... can we fix that ...)

pkg/config/load.go
```diff
@@ -0,0 +1,114 @@
...
+
+	if err := configSpec.Validate(); err != nil {
+		return nil, fmt.Errorf("invalid ci-operator config: %v", err)
+	}
+
+	return configSpec, nil
+}
+
+// path to ci-operator configuration file encodes information about tested code
+// .../$ORGANIZATION/$REPOSITORY/$BRANCH.$EXT
```
petr-muller: Total-nit: The comment of course rot. Let's update (it's now something like `...$ORG/$REPO/$ORG-$REPO-$BRANCH{,__$VARIANT}.yaml`) or delete it

pkg/config/load.go
```diff
@@ -0,0 +1,114 @@
...
+	Repo     string
+	Branch   string
+	Variant  string
+	Filename string
+}
+
+// We use the directory/file naming convention to encode useful information
+// about component repository information.
+// The convention for ci-operator config files in this repo:
+// ci-operator/config/ORGANIZATION/COMPONENT/BRANCH.yaml
```
petr-muller: Same

pkg/config/load.go
```diff
@@ -0,0 +1,114 @@
...
+func extractRepoElementsFromPath(configFilePath string) (*FilePathElements, error) {
+	configSpecDir := filepath.Dir(configFilePath)
+	repo := filepath.Base(configSpecDir)
+	if repo == "." || repo == "/" {
+		return nil, fmt.Errorf("could not extract repo from '%s' (expected path like '.../ORG/REPO/BRANCH.yaml", configFilePath)
```
petr-muller: Maybe un-bitrot the message too.

pkg/config/load.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	org := filepath.Base(filepath.Dir(configSpecDir))
+	if org == "." || org == "/" {
+		return nil, fmt.Errorf("could not extract org from '%s' (expected path like '.../ORG/REPO/BRANCH.yaml", configFilePath)
```
petr-muller: And here too.

smarterclayton:
I didn’t see anything that bothered me

smarterclayton:
Although I think we’re lacking a “release concept” that sits above promotion and defines some rules, but that can come later

stevekuznetsov:
@petr-muller tackled in #90

cmd/config-brancher/main.go
```diff
@@ -0,0 +1,99 @@
...
+		output.CommitTo(o.ConfigDir)
+	}
+}
+
+func generateBranchedConfigs(currentRelease, futureRelease string, input config.Info) []config.Info {
+	if !(promotion.PromotesOfficialImages(&input.Configuration) && input.Configuration.PromotionConfiguration.Name == currentRelease) {
+		return nil
+	}
+	input.Logger().Info("Branching configuration.")
+	// we need a deep copy and this is a simple albeit expensive hack to get there
```
petr-muller, line 68: We vendor `https://github.com/getlantern/deepcopy` in this repo which does basically the same. Maybe use it to shorten the code and avoid the need to explain why this is here?

cmd/config-brancher/main.go
```diff
@@ -0,0 +1,99 @@
...
+		// this config is the future promotion on the future branch
+		{Configuration: futureConfig, RepoInfo: copyInfoSwappingBranches(input.RepoInfo, futureBranchForFuturePromotion)},
+	}
+}
+
+func copyInfoSwappingBranches(input config.FilePathElements, newBranch string) config.FilePathElements {
+	intermediate := &input
+	output := *intermediate
+	output.Branch = newBranch
+	output.Filename = strings.Replace(output.Filename, input.Branch, newBranch, -1)
```
petr-muller: This looks a little fragile, it takes a single component that accidentally includes `master` in its name and we'll get surprising bugs.

Filename could be completely derived from the other elements (except the extension, but it could be taken from the old filename): `$org-$repo-$branch{,__$variant}.yaml`. Maybe it could even be a method instead of a member.

pkg/config/load.go
```diff
@@ -112,3 +113,36 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
...
+type Info struct {
+	Configuration cioperatorapi.ReleaseBuildConfiguration
+	RepoInfo      FilePathElements
+}
+
+func (i *Info) Logger() *logrus.Entry {
+	return LoggerForInfo(i.RepoInfo)
+}
+
+func (i *Info) CommitTo(dir string) {
```
petr-muller: I think this should propagate the errors and not eat them

pkg/config/load.go
```diff
@@ -112,3 +113,36 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
...
+	raw, err := yaml.Marshal(i.Configuration)
+	if err != nil {
+		i.Logger().WithError(err).Error("failed to marshal output CI Operator configuration")
+		return
+	}
+	outputFile := path.Join(
+		dir, i.RepoInfo.Org, i.RepoInfo.Repo,
+		fmt.Sprintf("%s-%s-%s.yaml", i.RepoInfo.Org, i.RepoInfo.Repo, i.RepoInfo.Branch),
```
petr-muller: 1. This ignores variants
2. `FilePathElements` has filename, why wasn't that used? This is probably also an opportunity to make that member a method instead of an attribute.

pkg/promotion/promotion.go
```diff
@@ -34,3 +39,64 @@ func extractPromotionName(configSpec *cioperatorapi.ReleaseBuildConfiguration) s
...
+	// futureBranchForFutureRelease is the branch that will promote to the future imagestream once we branch configs
+	var futureBranchForFutureRelease string
+	if currentBranch == "master" {
+		futureBranchForCurrentRelease = fmt.Sprintf("release-%s", currentRelease)
+		futureBranchForFutureRelease = currentBranch
+	} else if currentBranch == fmt.Sprintf("openshift-%s", currentRelease) {
+		futureBranchForCurrentRelease = currentBranch
+		futureBranchForFutureRelease = fmt.Sprintf("openshift-%s", futureRelease)
+	} else {
+		return "", "", fmt.Errorf("invalid branch %q promoting to current relesae", currentBranch)
```
petr-muller: nit:
```
current relesae
             ^^
```

cmd/config-brancher/main.go
```diff
@@ -0,0 +1,99 @@
...
+	"flag"
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/config"
+	"github.com/openshift/ci-operator-prowgen/pkg/promotion"
+	"github.com/openshift/ci-operator/pkg/api"
```
petr-muller: nit: separate local imports (`...-prowgen`) from 3rd party (`ci-operator`)

cmd/determinize-ci-operator/main.go
```diff
@@ -0,0 +1,55 @@
...
+
+import (
+	"flag"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/config"
+	"github.com/openshift/ci-operator-prowgen/pkg/promotion"
+	"github.com/openshift/ci-operator/pkg/api"
```
petr-muller: nit: separate local imports (`...-prowgen`) from 3rd party (`ci-operator`)

cmd/determinize-ci-operator/main.go
```diff
@@ -0,0 +1,55 @@
...
+	if err := o.Validate(); err != nil {
+		logrus.Fatalf("Invalid options: %v", err)
+	}
+
+	var toCommit []config.Info
+	if err := config.OperateOnCIOperatorConfigDir(o.ConfigDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.FilePathElements) error {
+		if (o.Org != "" && o.Org != repoInfo.Org) || (o.Repo != "" && o.Repo != repoInfo.Repo) {
+			return nil
+		}
+		if !(promotion.PromotesOfficialImages(configuration) && configuration.PromotionConfiguration.Name == o.CurrentRelease) {
```
petr-muller, line 36: I wonder whether a generically-sounding `determinize-*` tool should have this hard-coded. I understand this was actually needed only for those configs in this case, but maybe at least put it behind a flag?

petr-muller (commented):
I've reviewed everything except `cmd/repo-brancher/main.go` for which I really do not have any willpower left today :)

cmd/config-brancher/main.go
stevekuznetsov: Yeah I had that thought as well. Will change

cmd/determinize-ci-operator/main.go
stevekuznetsov: We only ever need to determinize configs we'd write to or generate, which this is the definition of. I don't see a reason to run the tool on other configs right now. In the future maybe we can make it more generic if we need it to be?

smarterclayton: I would say we don't need generic determinize right now - this is about getting ci-operator and prow configs right only.

smarterclayton:
This is really close, I think we can merge and follow up.  Getting the tooling ready is more important than being perfect on this.

petr-muller:
> This is really close, I think we can merge and follow up. Getting the tooling ready is more important than being perfect on this.

I don't mind, especially if it's urgent. Reviews just make future maintenance a bit easier :)

cmd/repo-brancher/main.go
```diff
@@ -0,0 +1,167 @@
...
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/ci-operator-prowgen/pkg/config"
+	"github.com/openshift/ci-operator-prowgen/pkg/promotion"
+	"github.com/openshift/ci-operator/pkg/api"
```
petr-muller: nit: separate local and vendor imports

cmd/repo-brancher/main.go
```diff
@@ -0,0 +1,167 @@
...
+			} else {
+				cmdLogger.WithFields(logrus.Fields{"output": string(out)}).Debug("Executed command.")
+				if string(out) == "" && !o.fastForward {
+					branchLogger.Info("Remote already has branch, skipping.")
+					continue
+				}
+			}
+
+			if !o.Confirm {
+				branchLogger.Info("Would create new branch.")
```
petr-muller, line 144: The message was a little confusing for me when I first tried to run the tool with `--fast-forward` but without `--confirm`, and I knew the branch should exist already.

cmd/repo-brancher/main.go
```diff
@@ -0,0 +1,167 @@
...
+		defer func() {
+			if err := os.Remove(tempDir); err != nil {
+				logrus.WithError(err).Fatal("Could not clean up temp dir for git operations")
+			}
+		}()
+		gitDir = tempDir
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(o.ConfigDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.FilePathElements) error {
+		logger := config.LoggerForInfo(*repoInfo)
```
petr-muller, line 80: This was a bit heavy to review, that anonymous method is quite long and could be broken down so some of the details are abstracted away.

petr-muller (commented):
Got to `repo-brancher` today, left few minor comments, but no showstopper.

stevekuznetsov:
OK, to shard the config I need  to merge this so we gen the correct prowjobs.

stevekuznetsov:
You have to import the world to load Prow plugin config, which is so amazing...

PR #91, opened 2019-02-25
Get template differences.

droslean:
This PR focus only on getting the templates differences. 
Also added a commit to add the template api, scheme libraries.


stevekuznetsov:
Seems like a reasonable start, but would love to see tests

droslean:
WIP -> working on tests

cmd/pj-rehearse/main.go
```diff
@@ -240,3 +255,65 @@ func getProwConfigs(candidatePath, baseSHA string) (*prowconfig.Config, *prowcon
...
+	templates := make(map[string]*templateapi.Template)
+	err := filepath.Walk(templatePath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return fmt.Errorf("prevent panic by handling failure accessing a path %q: %v", path, err)
+		}
+
+		if isYAML(path) {
+			contents, err := ioutil.ReadFile(path)
+			if err != nil {
+				return fmt.Errorf("could not read dir %s for template: %v", path, err)
```
petr-muller: The message `...could not read dir %s...` here is probably not correct, `path` is not a directory, but a YAML file.

cmd/pj-rehearse/main.go
```diff
@@ -240,3 +255,65 @@ func getProwConfigs(candidatePath, baseSHA string) (*prowconfig.Config, *prowcon
...
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error walking the path %q: %v", templatePath, err)
+	}
+	return templates, nil
+}
+
+func isYAML(file string) bool {
+	return filepath.Ext(file) == ".yaml"
```
petr-muller: My desire for correct code wants this to also allow `.yml` and rule out directories, but I'll understand if you do not want to bother :)

petr-muller (commented):
Minor feedback, otherwise fine, provided there are some tests.

While working on ci-operator config difference detection, I also needed to use the `git checkout` logic to obtain master & PR versions of the configs, and I also decided to duplicate it first. like you did. I was thinking about de-duplicate it later (make one method that would check out `BaseSHA`, load all the config, then check out PR SHA, load all the config...), and I presume you are thinking about it too. If so, I suggest to wait with it for my PR with the ci-operator config loading & comparison to not step on each other's toes.

cmd/pj-rehearse/main.go
droslean: addressed

cmd/pj-rehearse/main.go
droslean: addressed

pkg/diffs/diffs_test.go
```diff
@@ -148,3 +154,174 @@ func makeConfig(p []prowconfig.Presubmit) *prowconfig.Config {
...
+					Objects: baseObjects,
+				},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			before, after := testCase.getTemplates()
+			changedTemplates := GetChangedTemplates(before, after, logrus.NewEntry(logrus.New()))
```
stevekuznetsov: Can add test name as a `WithField` for better logging debug too

stevekuznetsov (commented):
So what are we doing with these differences now?

stevekuznetsov:
Having all code except the vendor in one commit would help review on this

pkg/diffs/diffs_test.go
droslean: addressed

droslean:
@stevekuznetsov I squash/update the PR with 2 commits. All the changes for the template differences with the tests included are now in one commit, and one commit for the vendors.

droslean:
> So what are we doing with these differences now?

Right now, we just print the changed templates. I want to focus on some cases in the next PRs
```
If template just changed/added ---> run a dummy job using that template with `TEST_COMMAND` `exit 0` 
If template changed/added and there is a changed job using that template -> run that job with the new template 
```

petr-muller (commented):
I was wondering for a little while whether we would need to test the `parseTemplates` code too, at least the bits that define what the method returns (the block on https://github.com/openshift/ci-operator-prowgen/pull/91/files#diff-06579b6b828bbff5df82617adb17e7bfR299) - something that would confirm that we are able to parse a template and put it into the map under the right key.

But I think it would be sufficient to put some template from `openshift/release` to ` test/pj-rehearse-integration/candidate/ci-operator/templates` - that would cause the code to be executed at least.

droslean:
> But I think it would be sufficient to put some template from `openshift/release` to ` test/pj-rehearse-integration/candidate/ci-operator/templates` - that would cause the code to be executed at least.

I want to update the integration tests to test the template data in a follow-up PR.

PR #96, opened 2019-02-27
Discover new/changed ci-operator configs

petr-muller:
Needed to implement a little `CompoundConfig` that loads all available
configs to memory, so that the comparison fits into our "load all config
from master, then load all config from PR" process.

As in https://github.com/openshift/ci-operator-prowgen/pull/91, we only discover the differences and log them. The followups are:

1. Optimize config loading so that we only do the `git checkout master; git checkout pr` dance once
2. Ditch the lazy ci-operator config inliner and start using the in-memory configs
3. Find jobs that use the changed configs and consider them for rehearsals.

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -112,3 +112,19 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
 		return nil
 	})
 }
+
+type CompoundCiopConfig map[string]*cioperatorapi.ReleaseBuildConfiguration
+
+func (compound *CompoundCiopConfig) add(handledConfig *cioperatorapi.ReleaseBuildConfiguration, handledElements *FilePathElements) error {
+	(*compound)[handledElements.Filename] = handledConfig
+	return nil
+}
+
+func CompoundLoad(path string) (CompoundCiopConfig, error) {
+	config := CompoundCiopConfig{}
+	if err := OperateOnCIOperatorConfigDir(path, config.add); err != nil {
+		return nil, err
+	}
+
+	return config, nil
+}
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -115,8 +115,8 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
 
 type CompoundCiopConfig map[string]*cioperatorapi.ReleaseBuildConfiguration
 
-func (compound *CompoundCiopConfig) add(handledConfig *cioperatorapi.ReleaseBuildConfiguration, handledElements *FilePathElements) error {
-	(*compound)[handledElements.Filename] = handledConfig
+func (compound CompoundCiopConfig) add(handledConfig *cioperatorapi.ReleaseBuildConfiguration, handledElements *FilePathElements) error {
+	compound[handledElements.Filename] = handledConfig
 	return nil
 }
 
--- a/pkg/diffs/diffs_test.go
+++ b/pkg/diffs/diffs_test.go
@@ -1,6 +1,7 @@
 package diffs
 
 import (
+	"reflect"
 	"testing"
 
 	"github.com/getlantern/deepcopy"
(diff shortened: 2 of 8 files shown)


pkg/config/load.go
```diff
@@ -112,3 +112,19 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
 		return nil
 	})
 }
+
+type CompoundCiopConfig map[string]*cioperatorapi.ReleaseBuildConfiguration
+
+func (compound *CompoundCiopConfig) add(handledConfig *cioperatorapi.ReleaseBuildConfiguration, handledElements *FilePathElements) error {
+	(*compound)[handledElements.Filename] = handledConfig
```
stevekuznetsov: I've never seen that one before... why do we need to deref?

pkg/diffs/diffs_test.go
```diff
@@ -3,14 +3,88 @@ package diffs
...
+		},
+	}}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			before, after := tc.configGenerator()
+			actual := GetChangedCiopConfigs(before, after, logrus.NewEntry(logrus.New()))
+			expected := tc.expected()
+
+			if !equality.Semantic.DeepEqual(expected, actual) {
```
stevekuznetsov: CI op configs are not k8s objects, should use reflect.DeepEqual?

test/pj-rehearse-integration/candidate/ci-operator/config/super/duper/super-duper-ciop-cfg-change.yaml
```diff
@@ -0,0 +1,28 @@
+base_images:
```
stevekuznetsov, line 1: Did the addition of this data impact the integration test at all?

stevekuznetsov (commented):
Couple questions, otherwise LGTM

pkg/config/load.go
droslean: I didn't check why the pointer needs to be resolved here, but If we need this I would prefer to have a constructor instead

pkg/config/load.go
```diff
@@ -112,3 +112,19 @@ func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi
 		return nil
 	})
 }
+
+type CompoundCiopConfig map[string]*cioperatorapi.ReleaseBuildConfiguration
+
+func (compound *CompoundCiopConfig) add(handledConfig *cioperatorapi.ReleaseBuildConfiguration, handledElements *FilePathElements) error {
```
droslean: Since there is not any error checking in this function, there is no reason to return an `error`, right?

pkg/diffs/diffs_test.go
droslean: But this is already working for non-k8s objects, right? If that so, I would like to keep it. 
Furthermore, if we use `reflect.DeepEqual` for the ci-op config differences, we will get a difference even if someone will change the order of the fields.

droslean (commented):
In general lgtm. Of course, we will do some follow-up PRS to merge the git checkout logic. Also to decide what to do with the changes

pkg/config/load.go
petr-muller: I was getting a `type *CompoundCiopConfig does not support indexing` error message. After some reading, it seems that one does not really need to use pointers to maps (if I understand it correctly, map variables are basically pointers by default), so I'll update the receiver.

pkg/diffs/diffs_test.go
petr-muller: I guess we should really use `reflect` here. Ordering should not matter as `CompoundConfig` is a map.

test/pj-rehearse-integration/candidate/ci-operator/config/super/duper/super-duper-ciop-cfg-change.yaml
petr-muller: Very little. It just causes the new code to be executed. So we catch panics and test only that current observable (sans new log messages) behavior does not change by the presence of ciop config and new code.

pkg/config/load.go
petr-muller: The method is used as the  `callback` param of `OperateOnCIOperatorConfigDir` method, so it needs to follow its signature.

pkg/config/load.go
droslean: Or there is no need to have a method at all. Just `make()` the map and pass it.

pkg/diffs/diffs_test.go
droslean: I mean the value of a specific key. But we should not care about that in the tests. So I agree to use `reflect` here.

petr-muller:
Addressed the feedback & rebased

PR #97, opened 2019-02-28
Add flag to derive JobSpec from local changes 

bbguimaraes:
/cc @droslean @petr-muller @stevekuznetsov

To be used in concert with the staging namespace for cases where we need to change more than the ProwJob. Only the manner in which we derive `JobSpec` is changed, so it should automatically pick up new additions (e.g. template tests) to the CI job.

Includes a few minor changes that didn't seem appropriate as separate PRs.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -40,6 +40,7 @@ func loadClusterConfig() (*rest.Config, error) {
 type options struct {
 	dryRun       bool
 	noFail       bool
+	local        bool
 	allowVolumes bool
 	debugLogPath string
 
@@ -53,6 +54,7 @@ func gatherOptions() options {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 	fs.BoolVar(&o.noFail, "no-fail", true, "Whether to actually end unsuccessfuly when something breaks")
+	fs.BoolVar(&o.local, "local", false, "Whether this is a local execution or part of a CI job")
 	fs.BoolVar(&o.allowVolumes, "allow-volumes", false, "Allows jobs with extra volumes to be rehearsed")
 
 	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
@@ -106,10 +108,17 @@ func main() {
 		gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
-	jobSpec, err := pjdwapi.ResolveSpecFromEnv()
-	if err != nil {
-		logrus.WithError(err).Error("could not read JOB_SPEC")
-		gracefulExit(o.noFail, misconfigurationOutput)
+	var jobSpec *pjdwapi.JobSpec
+	if o.local {
+		if jobSpec, err = config.NewLocalJobSpec(o.releaseRepoPath); err != nil {
+			logrus.WithError(err).Error("could not create local JobSpec")
+			gracefulExit(o.noFail, misconfigurationOutput)
+		}
+	} else {
+		if jobSpec, err = pjdwapi.ResolveSpecFromEnv(); err != nil {
+			logrus.WithError(err).Error("could not read JOB_SPEC")
+			gracefulExit(o.noFail, misconfigurationOutput)
+		}
 	}
 
 	prFields := logrus.Fields{prowgithub.OrgLogField: jobSpec.Refs.Org, prowgithub.RepoLogField: jobSpec.Refs.Repo}
@@ -169,7 +178,11 @@ func main() {
 		}
 	}
 
-	pjclient, err := rehearse.NewProwJobClient(clusterConfig, masterConfig.Prow.ProwJobNamespace, o.dryRun)
+	namespace := prConfig.Prow.ProwJobNamespace
+	if o.local {
+		namespace = "ci-stg"
+	}
+	pjclient, err := rehearse.NewProwJobClient(clusterConfig, namespace, o.dryRun)
 	if err != nil {
 		logger.WithError(err).Error("could not create a ProwJob client")
 		gracefulExit(o.noFail, misconfigurationOutput)
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -29,8 +29,8 @@ type ReleaseRepoConfig struct {
 	Templates  CiTemplates
 }
 
-func getCurrentSHA(repoPath string) (string, error) {
-	cmd := exec.Command("git", "rev-parse", "HEAD")
+func revParse(repoPath string, args ...string) (string, error) {
+	cmd := exec.Command("git", append([]string{"rev-parse"}, args...)...)
 	cmd.Dir = repoPath
 	sha, err := cmd.Output()
 	if err != nil {
@@ -85,7 +85,7 @@ func GetAllConfigs(releaseRepoPath string, logger *logrus.Entry) *ReleaseRepoCon
 // manipulations are propagated in the error return value. Errors occurred during the actual config loading are not
 // propagated, but the returned struct field will have a nil value in the appropriate field. The error is only logged.
 func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*ReleaseRepoConfig, error) {
-	currentSHA, err := getCurrentSHA(releaseRepoPath)
+	currentSHA, err := revParse(releaseRepoPath, "HEAD")
 	if err != nil {
 		return nil, fmt.Errorf("failed to get SHA of current HEAD: %v", err)
 	}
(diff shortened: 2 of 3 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -142,6 +178,7 @@ func main() {
 		}
 	}
 
+	defer git(o.candidatePath, []string{"checkout", jobSpec.Refs.Pulls[0].Ref})
```
droslean: why?

droslean (commented):
I like this because I found myself having a problem too when testing the tool locally.  The only thing that I don't like in this PR is that suddenly we do a lot of `git ` commands. I am starting to considering that maybe we need a library to handle all of that. Maybe `gopkg.in/libgit2/git2go.v25`?

cmd/pj-rehearse/main.go
bbguimaraes: It was annoying that the repository's `HEAD` ended up detached every time I executed it.

I haven't taken the time to figure out why yet.

bbguimaraes:
Base moved, will rebase.

cmd/pj-rehearse/main.go
bbguimaraes: That happens because we use SHAs instead of refs when we checkout. From `git`'s man page:

```
       --detach
           Rather than checking out a branch to work on it, check out a commit
           for inspection and discardable experiments. This is the default
           behavior of "git checkout <commit>" when <commit> is not a branch
           name.
```

While this doesn't matter in the CI job, it is annoying when executed in a local repository, specially when iterating on a job — the HEAD is left pointing to the right commit, but detached. Was there any reason to choose one rather than the other? If there was, I think the current solution of `defer`ring a `git checkout` to restore the original HEAD is fine.

bbguimaraes:
Rebased, added a flag to override the target namespace for ProwJobs (regular users cannot use the `ci` namespace, this allows them to `-n ci-stg`).

@petr-muller: you said you were rewriting the `git` calls. This touches them a bit, but it's a small change. I noticed I would have to change a few more things to avoid duplication after I rebased, but I can wait for your PR to merge to avoid unnecessary conflicts — maybe the changes then won't even be necessary.

cmd/pj-rehearse/main.go
```diff
@@ -114,10 +119,44 @@ func main() {
...
+			gracefulExit(o.noFail, misconfigurationOutput)
+		}
+		jobSpec = &pjdwapi.JobSpec{
+			Type: pjapi.PresubmitJob,
+			Refs: &pjapi.Refs{
+				Org:     "openshift",
+				Repo:    "release",
+				BaseSHA: baseSHA,
+				BaseRef: baseRef,
+				Pulls:   []pjapi.Pull{{SHA: pullSHA, Ref: pullRef}},
```
bbguimaraes: @stevekuznetsov: is anything horrible going to happen if we don't set the pull request number? We obviously have no value to set it to because these are local changes.

Nothing wrong happened in my tests with a PR number of `0`, but you surely know more of the dark corners of Prow.

cmd/pj-rehearse/main.go
petr-muller: > While this doesn't matter in the CI job, it is annoying when executed in a local repository, specially when iterating on a job — the HEAD is left pointing to the right commit, but detached. Was there any reason to choose one rather than the other? 

I think yes. My understanding is that when you run in a decorated Prowjob, you do not have `jobSpec.Refs.Pulls[0].Ref` checked out, you have it merged with the base. I'm not sure if the implementation does it in a way where the ref name is moved to the merge commit sha.

petr-muller:
> you said you were rewriting the `git` calls. This touches them a bit, but it's a small change. I noticed I would have to change a few more things to avoid duplication after I rebased, but I can wait for your PR to merge to avoid unnecessary conflicts — maybe the changes then won't even be necessary.

:+1: my intention is to deduplicate the git checkout logic from the three `getX` methods, hopefully those are the parts where you spotted the duplication. Unfortunately it's not a stupid simple change, there are some quirks in error handling that need to be thought out well (some configs are important to load, some not that much).

cmd/pj-rehearse/main.go
bbguimaraes: Right, looking at the code, it is all done on top of the base ref.

I think `defer`ing a checkout is the simplest solution here. I have moved it inside the `if o.local` conditional, since that is not necessary in CI.

cmd/pj-rehearse/main.go
stevekuznetsov: I think that should be fine

cmd/pj-rehearse/main.go
stevekuznetsov: Creating named refs gets real messy real quick in CI -- you have to worry about collisions, etc. The CI knows exactly what it wants to be doing with SHAs and can do it more cleanly that way.

cmd/pj-rehearse/main.go
```diff
@@ -61,8 +63,11 @@ func gatherOptions() options {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually submit rehearsal jobs to Prow")
 	fs.BoolVar(&o.noFail, "no-fail", true, "Whether to actually end unsuccessfuly when something breaks")
+	fs.BoolVar(&o.local, "local", false, "Whether this is a local execution or part of a CI job")
 	fs.BoolVar(&o.allowVolumes, "allow-volumes", false, "Allows jobs with extra volumes to be rehearsed")
 
+	fs.StringVar(&o.namespace, "n", "", "Short for --namespace")
```
stevekuznetsov: Will anyone ever set this to something other than `ci-stg`? Why expose it to the user?

cmd/pj-rehearse/main.go
```diff
@@ -114,10 +119,45 @@ func main() {
...
-		logrus.WithError(err).Error("could not read JOB_SPEC")
-		gracefulExit(o.noFail, misconfigurationOutput)
+	var jobSpec *pjdwapi.JobSpec
+	if o.local {
+		pullRef, err := git(o.candidatePath, []string{"rev-parse", "--abbrev-ref", "HEAD"})
+		if err != nil {
+			logrus.WithError(err).Error("could not get current branch")
+			gracefulExit(o.noFail, misconfigurationOutput)
+		}
+		baseRef, err := git(o.candidatePath, []string{"rev-parse", "--abbrev-ref", pullRef + "@{upstream}"})
```
stevekuznetsov: I imagine people will not always have dev workflows that work with this. I think all of Prow is well-formed to use the SHA whenever present and fall back to the ref if not, so we _could_ set these to dummy values or at least be best-effort about them. Let's see if people end up erroring on this before we change anything.

stevekuznetsov (approved):
These changes seem reasonable to me

cmd/pj-rehearse/main.go
bbguimaraes: Probably not (we do have some creative users), but I still think it is a nice option to have.

How about we keep the flag but default to `ci-stg` when `o.local`?

cmd/pj-rehearse/main.go
bbguimaraes: I went with the idea of an upstream branch because it is a concept that is native to git, but I am not crazy about it and am open to suggestions. Some others I considered:

- `master`: why not?
- a separate flag: sounds like too much work.
- some other git magic I am no aware of?

cmd/pj-rehearse/main.go
droslean: This should be empty only in a `periodic` job, right? Since that we rehearse only `presubmits` and also we populate the `jobspec` we won't have any problem with that.

bbguimaraes: This is the fake `JobSpec` that is constructed based on local changes, where we don't have a pull request. The CI code path is unchanged.

cmd/pj-rehearse/main.go
stevekuznetsov: Creative users cause us headaches. Let's not expose this until we have a clear need, right now I think we simply

```go
namespace := "ci"
if o.runLocal {
    namespace = "ci-stg"
}
```

or whatever

cmd/pj-rehearse/main.go
stevekuznetsov: Let's use `master` if the user has not set an `upstream`? We don't have any other upstream branches on the `release` repo anyway right?

stevekuznetsov (approved):

bbguimaraes:
- rebased
- `master` is used when the branch doesn't have an upstream
- removed the `-n/--namespace` flag, `ci-stg` is used when `--local`
- the branch name (if it exists) is used to restore the original revision — it was easier and cleaner to implement it this way

stevekuznetsov (commented):
/lgtm
/approve

PR #99, opened 2019-03-04
 rehearse: optimize git manipulation when loading configs

petr-muller:
Previously we loaded all three CI config fragments (Prow config,
ci-operator config, templates) separately. Because we need to load the
configs from two different revisions of the release repo, all three
loading methods handled the "check out master revision, load config,
check out tested revision" procedure themselves, which resulted in the
whole operation being performed three times unnecessarily. This was
optimized so that we separate loading of all config from the git
operation.

A little tricky area is the error handling and propagation. The errors
occurred during config loading itself are not propagated via error values
but by the appropriate field of the config having a nil value. This is
because we need granular information about which configs loaded
correctly and which not, so the tool can decide a right course of
action. The decisions are as follows:

- Exit when any Prow config version loading failed, because we need to
  compare both versions to compute which jobs should be rehearsed.
- Exit when PR version of templates or ci-operator config loading
  failed, because we need these versions to provide them to rehearsed
  jobs.
- Do not compare templates when master version of templates fails to
  load.
- Do not compare ci-operator config when master version of these configs
  fails to load.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -51,8 +51,8 @@ type options struct {
 	allowVolumes bool
 	debugLogPath string
 
-	candidatePath  string
-	rehearsalLimit int
+	releaseRepoPath string
+	rehearsalLimit  int
 }
 
 func gatherOptions() options {
@@ -64,7 +64,7 @@ func gatherOptions() options {
 	fs.BoolVar(&o.allowVolumes, "allow-volumes", false, "Allows jobs with extra volumes to be rehearsed")
 
 	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
-	fs.StringVar(&o.candidatePath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
+	fs.StringVar(&o.releaseRepoPath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
 
 	fs.IntVar(&o.rehearsalLimit, "rehearsal-limit", 15, "Upper limit of jobs attempted to rehearse (if more jobs would be rehearsed, none will)")
 
@@ -73,7 +73,7 @@ func gatherOptions() options {
 }
 
 func validateOptions(o options) error {
-	if len(o.candidatePath) == 0 {
+	if len(o.releaseRepoPath) == 0 {
 		return fmt.Errorf("--candidate-path was not provided")
 	}
 	return nil
@@ -143,13 +143,13 @@ func main() {
 		}
 	}
 
-	prowConfig, prowPRConfig, err := getProwConfigs(o.candidatePath, jobSpec.Refs.BaseSHA)
+	prowConfig, prowPRConfig, err := getProwConfigs(o.releaseRepoPath, jobSpec.Refs.BaseSHA)
 	if err != nil {
 		logger.WithError(err).Error("could not load prow configs")
 		gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
-	ciopConfig, ciopPRConfig, err := getCiopConfigs(o.candidatePath, jobSpec.Refs.BaseSHA)
+	ciopConfig, ciopPRConfig, err := getCiopConfigs(o.releaseRepoPath, jobSpec.Refs.BaseSHA)
 	if err != nil {
 		logger.WithError(err).Error("could not load ci-operator configs")
 		gracefulExit(o.noFail, misconfigurationOutput)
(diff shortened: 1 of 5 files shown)


stevekuznetsov (commented):
What was driving the error propagation behavior? If we fail to load any part of the config from either revision, we can't do our job. right?

petr-muller:
> What was driving the error propagation behavior? If we fail to load any part of the config from either revision, we can't do our job. right?

That's not *entirely* true. We do not need the **master** versions for anything else than to serve as a base for comparison, but at the same time, we have three such bases, so we can ignore when only some of them fail to load. For example, when we fail to load master ci-operator configs, we cannot compare them and find out which jobs are affected, but we can rehearse jobs that were changed directly.

We need the PR version of everything to load correctly, because we need to inline them to rehearsed jobs. Also, if something fails there, the PR is likely to fail other checks anyway.

pkg/config/template.go
```diff
@@ -0,0 +1,49 @@
...
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	templateapi "github.com/openshift/api/template/v1"
+	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
+)
+
+type CiTemplates map[string]*templateapi.Template
```
droslean, line 14: nit: I would prefer to be named just as `Templates`.

cmd/pj-rehearse/main.go
```diff
@@ -143,29 +135,46 @@ func main() {
...
-		logger.WithError(err).Error("could not load prow configs")
+		logger.WithError(err).Error("could not load configuration from base revision of release repo")
 		gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
-	ciopConfig, ciopPRConfig, err := getCiopConfigs(o.candidatePath, jobSpec.Refs.BaseSHA)
-	if err != nil {
-		logger.WithError(err).Error("could not load ci-operator configs")
+	// We always need both Prow config versions, otherwise we cannot compare them
+	if masterConfig.Prow == nil || prConfig.Prow == nil {
```
droslean, line 146: can we do that inside `GetAllConfigs`?

cmd/pj-rehearse/main.go
```diff
@@ -143,29 +135,46 @@ func main() {
...
+		logger.WithError(err).Error("could not load Prow configs from base or tested revision of release repo")
+		gracefulExit(o.noFail, misconfigurationOutput)
+	}
+	// We always need PR versions of templates and ciop config, otherwise we cannot provide them to rehearsed jobs
+	if prConfig.Templates == nil || prConfig.CiOperator == nil {
```
droslean, line 151: this one too.

cmd/pj-rehearse/main.go
```diff
@@ -143,29 +135,46 @@ func main() {
...
+	// We can only detect changes if we managed to load both ci-operator config versions
+	if masterConfig.CiOperator != nil && prConfig.CiOperator != nil {
+		changedCiopConfigs := diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
+		for name := range changedCiopConfigs {
+			logger.WithField("ciop-config", name).Info("Changed ci-operator config")
+		}
+	}
+
+	// We can only detect changes if we managed to load both CI template versions
+	if masterConfig.Templates != nil && prConfig.Templates != nil {
```
droslean, line 165: this can be checked also in `GetAllConfigs`.

droslean (commented):
In general, this PR looks good to me. I only commented on some nits.

cmd/pj-rehearse/main.go
petr-muller: Not really. The decision to `os.Exit` should be certainly kept at the `main()` level, library methods should not `Exit`. Returning a single `error` value with the _at least one thing failed_ semantic was not a good option because we can handle some failures differently (see https://github.com/openshift/ci-operator-prowgen/pull/99#issuecomment-469331170). Hence this: the contract is quite simple (_load all configs, those that fail will be `nil`_) and the callsite knows better what members it needs for what.

pkg/config/template.go
petr-muller: Do we have some templates that aren't used for CI? If so then I'll rename, otherwise I prefer the more specific name.

cmd/pj-rehearse/main.go
petr-muller: This is exactly the place that cannot - `GetAllConfigs` only sees a single revision of the config, and it does not know anything about comparison of templates.

PR #100, opened 2019-03-05
pj-rehearse - Add configMaps logic on changed templates

droslean:
This PR adds the logic of the creation of temporary configMaps for the changed templates.

- [X] Create configMaps for all the changed templates
- [X] Cleanup all the temporary configMaps after the finish.
- [X] Detect if the jobs are using the changed template and use the temporary configMap instead.
- [X] Support this procedure only in dry-run false.
- [X] Add tests

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -100,6 +100,15 @@ func gracefulExit(suppressFailures bool, message string) {
 	os.Exit(1)
 }
 
+func gracefulExitWithConfigMapCleanup(suppressFailures bool, message string, cmManager *config.TemplateCMManager) {
+	if cmManager != nil {
+		if err := cmManager.CleanupCMTemplates(); err != nil {
+			fmt.Fprintln(os.Stderr, "Failed to cleanup temporary configmaps")
+		}
+	}
+	gracefulExit(suppressFailures, message)
+}
+
 func main() {
 	o := gatherOptions()
 	err := validateOptions(o)
@@ -191,13 +200,13 @@ func main() {
 	cmManager := config.NewTemplateCMManager(cmClient, prNumber, logger, changedTemplates)
 	if err := cmManager.CreateCMTemplates(); err != nil {
 		logger.WithError(err).Error("couldn't create template configMap")
-		gracefulExit(o.noFail, "")
+		gracefulExitWithConfigMapCleanup(o.noFail, "", cmManager)
 	}
 
 	pjclient, err := rehearse.NewProwJobClient(clusterConfig, namespace, o.dryRun)
 	if err != nil {
 		logger.WithError(err).Error("could not create a ProwJob client")
-		gracefulExit(o.noFail, misconfigurationOutput)
+		gracefulExitWithConfigMapCleanup(o.noFail, misconfigurationOutput, cmManager)
 	}
 
 	debugLogger := logrus.New()
@@ -208,7 +217,7 @@ func main() {
 			debugLogger.Out = f
 		} else {
 			logger.WithError(err).Error("could not open debug log file")
-			gracefulExit(o.noFail, "")
+			gracefulExitWithConfigMapCleanup(o.noFail, "", cmManager)
 		}
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
@@ -219,25 +228,29 @@ func main() {
 	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
-		gracefulExit(true, "")
+		gracefulExitWithConfigMapCleanup(true, "", cmManager)
 	} else if len(rehearsals) > o.rehearsalLimit {
 		jobCountFields := logrus.Fields{
 			"rehearsal-threshold": o.rehearsalLimit,
 			"rehearsal-jobs":      len(rehearsals),
 		}
 		logger.WithFields(jobCountFields).Info("Would rehearse too many jobs, will not proceed")
-		gracefulExit(true, "")
+		gracefulExitWithConfigMapCleanup(true, "", cmManager)
 	}
 
 	executor := rehearse.NewExecutor(rehearsals, prNumber, o.releaseRepoPath, jobSpec.Refs, o.dryRun, loggers, pjclient)
 	success, err := executor.ExecuteJobs()
 	if err != nil {
 		logger.WithError(err).Error("Failed to rehearse jobs")
-		gracefulExit(o.noFail, rehearseFailureOutput)
+		gracefulExitWithConfigMapCleanup(o.noFail, rehearseFailureOutput, cmManager)
 	}
 	if !success {
 		logger.Error("Some jobs failed their rehearsal runs")
-		gracefulExit(o.noFail, jobsFailureOutput)
+		gracefulExitWithConfigMapCleanup(o.noFail, jobsFailureOutput, cmManager)
 	}
 	logger.Info("All jobs were rehearsed successfully")
+
+	if err := cmManager.CleanupCMTemplates(); err != nil {
+		logger.WithError(err).Error("Failed to cleanup temporary configmaps")
+	}
 }
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -168,9 +168,10 @@ func main() {
 		changedCiopConfigs = diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
 	}
 
+	var changedTemplates config.CiTemplates
 	// We can only detect changes if we managed to load both CI template versions
 	if masterConfig.Templates != nil && prConfig.Templates != nil {
-		changedTemplates := diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
+		changedTemplates = diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
 		for name := range changedTemplates {
 			logger.WithField("template-name", name).Info("Changed template")
 		}
@@ -202,7 +203,7 @@ func main() {
 	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
 	toRehearse.AddAll(diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger))
 
-	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes)
+	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		os.Exit(0)
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -181,6 +181,19 @@ func main() {
 	if o.local {
 		namespace = "ci-stg"
 	}
+
+	cmClient, err := rehearse.NewCMClient(clusterConfig, namespace, o.dryRun)
+	if err != nil {
+		logger.WithError(err).Error("could not create a configMap client")
+		gracefulExit(o.noFail, misconfigurationOutput)
+	}
+
+	cmManager := config.NewTemplateCMManager(cmClient, prNumber, logger, changedTemplates)
+	if err := cmManager.CreateCMTemplates(); err != nil {
+		logger.WithError(err).Error("couldn't create template configMap")
+		gracefulExit(o.noFail, "")
+	}
+
 	pjclient, err := rehearse.NewProwJobClient(clusterConfig, namespace, o.dryRun)
 	if err != nil {
 		logger.WithError(err).Error("could not create a ProwJob client")
@@ -206,14 +219,14 @@ func main() {
 	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
-		os.Exit(0)
+		gracefulExit(true, "")
 	} else if len(rehearsals) > o.rehearsalLimit {
 		jobCountFields := logrus.Fields{
 			"rehearsal-threshold": o.rehearsalLimit,
 			"rehearsal-jobs":      len(rehearsals),
 		}
 		logger.WithFields(jobCountFields).Info("Would rehearse too many jobs, will not proceed")
-		os.Exit(0)
+		gracefulExit(true, "")
 	}
 
 	executor := rehearse.NewExecutor(rehearsals, prNumber, o.releaseRepoPath, jobSpec.Refs, o.dryRun, loggers, pjclient)
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -168,7 +168,7 @@ func main() {
 		changedCiopConfigs = diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
 	}
 
-	var changedTemplates config.CiTemplates
+	changedTemplates := make(config.CiTemplates)
 	// We can only detect changes if we managed to load both CI template versions
 	if masterConfig.Templates != nil && prConfig.Templates != nil {
 		changedTemplates = diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -2,6 +2,8 @@ package config
 
 import (
 	"bytes"
+	"crypto/sha256"
+	"encoding/base32"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -90,17 +92,13 @@ func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logg
 // CreateCMTemplates creates configMaps for all the changed templates.
 func (c *TemplateCMManager) CreateCMTemplates() error {
 	var errors []error
-	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
-
 	for filename, template := range c.templates {
-		buf := new(bytes.Buffer)
-
-		err := s.Encode(template, buf)
+		templateData, err := GetTemplateData(template)
 		if err != nil {
 			errors = append(errors, err)
 		}
 
-		cmName := GetTempCMName(c.prNumber, template.Name)
+		cmName := GetTempCMName(template.Name, filename, templateData)
 		cm := &v1.ConfigMap{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: cmName,
@@ -109,7 +107,7 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
 					rehearseLabelPull: strconv.Itoa(c.prNumber),
 				},
 			},
-			Data: map[string]string{filename: buf.String()},
+			Data: map[string]string{filename: templateData},
 		}
 
 		c.logger.WithFields(logrus.Fields{"template-name": template.Name, "cm-name": cmName}).Info("creating rehearsal configMap for template")
@@ -139,7 +137,39 @@ func (c *TemplateCMManager) CleanupCMTemplates() error {
 	return nil
 }
 
-// GetTempCMName converts a template name to a temporary name including the pr number.
-func GetTempCMName(prNumber int, templateName string) string {
-	return fmt.Sprintf("rehearse-%d-%s", prNumber, templateName)
+func GetTempCMName(templateName, filename, templateData string) string {
+	inputs := []string{filename, templateData}
+	return fmt.Sprintf("rehearse-%s-%s", inputHash(inputs), templateName)
+}
+
+func GetTemplateData(template *templateapi.Template) (string, error) {
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+	buf := new(bytes.Buffer)
+
+	err := s.Encode(template, buf)
+	if err != nil {
+		return "", err
+	}
+
+	return buf.String(), nil
+}
+
+// oneWayEncoding can be used to encode hex to a 62-character set (0 and 1 are duplicates) for use in
+// short display names that are safe for use in kubernetes as resource names.
+var oneWayNameEncoding = base32.NewEncoding("bcdfghijklmnpqrstvwxyz0123456789").WithPadding(base32.NoPadding)
+
+// inputHash returns a string that hashes the unique parts of the input to avoid collisions.
+func inputHash(inputs []string) string {
+	hash := sha256.New()
+
+	// the inputs form a part of the hash
+	for _, s := range inputs {
+		hash.Write([]byte(s))
+	}
+
+	// Object names can't be too long so we truncate
+	// the hash. This increases chances of collision
+	// but we can tolerate it as our input space is
+	// tiny.
+	return oneWayNameEncoding.EncodeToString(hash.Sum(nil)[:5])
 }
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -1,17 +1,38 @@
 package config
 
 import (
+	"bytes"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
+
+	"github.com/sirupsen/logrus"
 
 	templateapi "github.com/openshift/api/template/v1"
 	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
+
+	"k8s.io/api/core/v1"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	kutilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"k8s.io/client-go/kubernetes/scheme"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 )
 
+// CiTemplates is a map of all the changed templates
 type CiTemplates map[string]*templateapi.Template
+
+const (
+	createByRehearse  = "created-by-pj-rehearse"
+	rehearseLabelPull = "ci.openshift.org/rehearse-pull"
+)
 
 func getTemplates(templatePath string) (CiTemplates, error) {
 	templates := make(map[string]*templateapi.Template)
@@ -46,4 +67,78 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 
 func isYAML(file string, info os.FileInfo) bool {
 	return !info.IsDir() && (filepath.Ext(file) == ".yaml" || filepath.Ext(file) == ".yml")
+}
+
+// TemplateCMManager holds the details needed for the configmap controller
+type TemplateCMManager struct {
+	cmclient  corev1.ConfigMapInterface
+	prNumber  int
+	logger    *logrus.Entry
+	templates CiTemplates
+}
+
+// NewTemplateCMManager creates a new TemplateCMManager
+func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logger *logrus.Entry, templates CiTemplates) *TemplateCMManager {
+	return &TemplateCMManager{
+		cmclient:  cmclient,
+		prNumber:  prNumber,
+		logger:    logger,
+		templates: templates,
+	}
+}
+
+// CreateCMTemplates creates configMaps for all the changed templates.
+func (c *TemplateCMManager) CreateCMTemplates() error {
+	var errors []error
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	for filename, template := range c.templates {
+		buf := new(bytes.Buffer)
+
+		err := s.Encode(template, buf)
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		cmName := getTempCMName(c.prNumber, template.Name)
+		cm := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cmName,
+				Labels: map[string]string{
+					createByRehearse:  "true",
+					rehearseLabelPull: strconv.Itoa(c.prNumber),
+				},
+			},
+			Data: map[string]string{filename: buf.String()},
+		}
+
+		c.logger.WithFields(logrus.Fields{"template-name": template.Name, "cm-name": cmName}).Info("creating rehearsal configMap for template")
+		if _, err := c.cmclient.Create(cm); err != nil {
+			if kerrors.IsAlreadyExists(err) {
+				if _, err := c.cmclient.Update(cm); err != nil {
+					errors = append(errors, fmt.Errorf("could not update existing configmap: %v", err))
+				}
+			} else {
+				errors = append(errors, err)
+			}
+		}
+	}
+	return kutilerrors.NewAggregate(errors)
+}
+
+// CleanupCMTemplates deletes all the configMaps that have been created for the changed templates.
+func (c *TemplateCMManager) CleanupCMTemplates() error {
+	c.logger.Info("deleting temporary template configMaps")
+	if err := c.cmclient.DeleteCollection(&metav1.DeleteOptions{},
+		metav1.ListOptions{LabelSelector: fields.Set{
+			createByRehearse:  "true",
+			rehearseLabelPull: strconv.Itoa(c.prNumber),
+		}.AsSelector().String()}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func getTempCMName(prNumber int, templateName string) string {
+	return fmt.Sprintf("rehearse-%d-%s", prNumber, templateName)
 }
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -139,6 +139,7 @@ func (c *TemplateCMManager) CleanupCMTemplates() error {
 	return nil
 }
 
+// GetTempCMName converts a template name to a temporary name including the pr number.
 func GetTempCMName(prNumber int, templateName string) string {
 	return fmt.Sprintf("rehearse-%d-%s", prNumber, templateName)
 }
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -100,7 +100,7 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
 			errors = append(errors, err)
 		}
 
-		cmName := getTempCMName(c.prNumber, template.Name)
+		cmName := GetTempCMName(c.prNumber, template.Name)
 		cm := &v1.ConfigMap{
 			ObjectMeta: metav1.ObjectMeta{
 				Name: cmName,
@@ -139,6 +139,6 @@ func (c *TemplateCMManager) CleanupCMTemplates() error {
 	return nil
 }
 
-func getTempCMName(prNumber int, templateName string) string {
+func GetTempCMName(prNumber int, templateName string) string {
 	return fmt.Sprintf("rehearse-%d-%s", prNumber, templateName)
 }
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -32,7 +32,7 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 						template.Name = filepath.Base(path)
 						template.Name = strings.TrimSuffix(template.Name, filepath.Ext(template.Name))
 					}
-					templates[template.Name] = template
+					templates[filepath.Base(path)] = template
 				}
 			}
 		}
--- a/pkg/config/template_test.go
+++ b/pkg/config/template_test.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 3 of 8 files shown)


droslean:
WIP - fixing tests.

pkg/config/template.go
```diff
@@ -47,3 +60,60 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+
+// CreateCMTemplates creates configMaps for all the changed templates.
+func (c *TemplateCMController) CreateCMTemplates() []error {
+	var errors []error
+	s := json.NewYAMLSerializer(json.DefaultMetaFactory, scheme.Scheme, scheme.Scheme)
+
+	for filename, template := range c.templates {
+		buf := new(bytes.Buffer)
+
+		// Encode the object to YAML.
```
petr-muller: Don't think the comment is that helpful here, the situation is not that complicated.

pkg/config/template.go
```diff
@@ -47,3 +60,60 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		err := s.Encode(template, buf)
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		cm := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("rehearse-%d-%s", c.prNumber, template.Name)},
```
petr-muller: Consider extracting the name computation to a method. Line 114 does the exact same thing, if we ever need to change the names, let's have a single place to change.

pkg/rehearse/jobs.go
```diff
@@ -185,13 +187,34 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
-func ConfigureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool) []*prowconfig.Presubmit {
+func ConfigureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool, cmClient coreclientset.ConfigMapInterface, templates map[string]*templateapi.Template, dryRun bool) []*prowconfig.Presubmit {
 	rehearsals := []*prowconfig.Presubmit{}
 
 	rehearsalsFiltered := filterJobs(toBeRehearsed, allowVolumes, loggers.Job)
 	for repo, jobs := range rehearsalsFiltered {
 		for _, job := range jobs {
 			jobLogger := loggers.Job.WithFields(logrus.Fields{"target-repo": repo, "target-job": job.Name})
+
+			if !dryRun {
```
petr-muller: I don't think we need to avoid doing this while in dry mode. The code only reads from the remote system, there is no modifying call.

pkg/rehearse/jobs.go
```diff
@@ -185,13 +187,34 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
+				// check if the job has a volumeSource from a configmap
+				// then check if that configmap has a key same as a changed template
+				// and if so, change the name to the temporary format (rehearse-$pr-$templatename)
+				// example: rehearse-2715-cluster-launch-e2e-openshift-ansible
+				for index, volume := range job.Spec.Volumes {
+					if volume.VolumeSource.ConfigMap != nil {
+						cm, err := cmClient.Get(volume.VolumeSource.ConfigMap.Name, metav1.GetOptions{})
+						if err != nil {
+							fmt.Errorf("%v", err)
```
petr-muller: This looks like an unfinished error handling.

pkg/rehearse/jobs.go
```diff
@@ -185,13 +187,34 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
+						}
+						for key := range cm.Data { // It should be only one.
+							if _, ok := templates[key]; ok {
+								job.Spec.Volumes[index].VolumeSource.ConfigMap.Name = fmt.Sprintf("rehearse-%d-%s", prNumber, volume.VolumeSource.ConfigMap.Name)
```
petr-muller: Here's an additional configmap name computation.

pkg/rehearse/jobs.go
```diff
@@ -185,13 +187,34 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
 	for repo, jobs := range rehearsalsFiltered {
 		for _, job := range jobs {
 			jobLogger := loggers.Job.WithFields(logrus.Fields{"target-repo": repo, "target-job": job.Name})
+
+			if !dryRun {
+				// check if the job has a volumeSource from a configmap
+				// then check if that configmap has a key same as a changed template
+				// and if so, change the name to the temporary format (rehearse-$pr-$templatename)
+				// example: rehearse-2715-cluster-launch-e2e-openshift-ansible
+				for index, volume := range job.Spec.Volumes {
```
petr-muller: I think this code needs to be:

1. performed over the actual rehearsal, not the source job. `makeRehearsalPresubmit` makes a copy with the basic set of needed modifications - that should happen first. Further modifications, like this one, needs to happen on that object, if only to avoid modifying the objects passed as input.
2. extracted to a method so that this method stays clean and on a single level of abstraction

pkg/rehearse/jobs.go
```diff
@@ -185,13 +187,34 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
...
+					if volume.VolumeSource.ConfigMap != nil {
+						cm, err := cmClient.Get(volume.VolumeSource.ConfigMap.Name, metav1.GetOptions{})
```
petr-muller: This is not really a code comment but a question: I'm having a hard time wrapping my head around why this is even needed, could you elaborate a bit on why do we need to reach to the cluster to check what's in the configmap? Isn't everything needed present in the job config?

cmd/pj-rehearse/main.go
```diff
@@ -161,14 +164,42 @@ func main() {
 		}
 	}
 
+	changedTemplates := make(map[string]*templateapi.Template)
```
petr-muller: I wanted to suggest this should be `config.CiTemplates`, but now I see I haven't fixed those types in `diffs.GetChangedTemplates` too :)

cmd/pj-rehearse/main.go
```diff
@@ -161,14 +164,42 @@ func main() {
...
 	if masterConfig.Templates != nil && prConfig.Templates != nil {
-		changedTemplates := diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
+		changedTemplates = diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
 		for name := range changedTemplates {
 			logger.WithField("template-name", name).Info("Changed template")
 		}
 	}
 
+	var cmClient corev1.ConfigMapInterface
+	if !o.dryRun {
```
petr-muller: This does not need to be guarded in dry-run mode - we have a client that handles dry-run itself.

cmd/pj-rehearse/main.go
```diff
@@ -161,14 +164,42 @@ func main() {
...
+			logger.WithError(err).Error("could not create a configMap client")
+			gracefulExit(o.noFail, misconfigurationOutput)
+		}
+
+		cmController := config.NewTemplateCMController(cmClient, prNumber, logger, changedTemplates)
+		if errors := cmController.CreateCMTemplates(); len(errors) > 0 {
+			for _, err := range errors {
+				logger.WithError(err).Error("couldn't create template configMap")
+			}
+			gracefulExit(o.noFail, "")
```
petr-muller: I think the cleanup here is a little more complicated - `CreateCMTemplates` might have created few CMs successfully before a failure occurred, so we would leave thes on the cluster.

cmd/pj-rehearse/main.go
```diff
@@ -161,14 +164,42 @@ func main() {
...
+
+		}
+
+		defer func() {
```
petr-muller: I don't think this will work properly - IIRC `defer` does not run on `os.Exit`, right? And we use that a lot for failure modes...

pkg/config/template.go
petr-muller: Can we also label the CMs, just like we label the rehearsal jobs.

petr-muller (commented):
Good start! I left some comments for further iterations, and we'll obviously need tests, but it's a solid base for the feature.

One thing that we'll need to take care of is handling the CM lifecycle properly in the presence of retests and reruns. The sad fact about rehearsals is that Prow itself never terminates the rehearsal jobs, for example when someone pushes new commits to a PR while the rehearsals are running. The reason for this is the same as in other instances - Prow only takes jobs configured for a repo into the account, and rehearsals are not configured for a repo.

This means that when this code runs, it's possible that the CM with the same name (we shard the CM names only by PR number) already exists and is used by some jobs that are running. I'm not 100% sure about the implications of this but we should take that into account.

I think we should collect all existing rehearse resources for the given PR at the beginning (labeling should help us with this) and terminate / delete them, but this should probably come in a separate PR.

cmd/pj-rehearse/main.go
droslean: Good point. So we should ignore `doesn't exist` errors

cmd/pj-rehearse/main.go
droslean: I think we should cleanup the `configMaps` inside the `gracefulExit`.

pkg/config/template.go
droslean: that could be useful when we delete all the configmaps. If we delete them by label, we won't have any forgotten cm

cmd/pj-rehearse/main.go
stevekuznetsov: We should use https://golang.org/pkg/runtime/#Goexit and defer `os.Exit`

pkg/config/template.go
```diff
@@ -47,3 +60,60 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		c.logger.WithField("template-name", template.Name).Info("creating rehearsal configMap for template")
+		if _, err := c.cmclient.Create(cm); err != nil && !kerrors.IsAlreadyExists(err) {
+			errors = append(errors, err)
+		}
+	}
+	return errors
+}
+
+// CleanupCMTemplates deletes all the configMaps that have been created for the changed templates.
+func (c *TemplateCMController) CleanupCMTemplates() []error {
```
stevekuznetsov: This would be better using a label selector, makes it not possible for us to lose track of one

pkg/config/template.go
```diff
@@ -47,3 +60,60 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+	return &TemplateCMController{
+		cmclient:  cmclient,
+		prNumber:  prNumber,
+		logger:    logger,
+		templates: templates,
+	}
+}
+
+// CreateCMTemplates creates configMaps for all the changed templates.
+func (c *TemplateCMController) CreateCMTemplates() []error {
```
stevekuznetsov: Might want to return an aggregate single error, unless you want the caller to do something with a specific error from this slice

pkg/config/template.go
```diff
@@ -32,7 +45,7 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 						template.Name = filepath.Base(path)
 						template.Name = strings.TrimSuffix(template.Name, filepath.Ext(template.Name))
 					}
-					templates[template.Name] = template
+					templates[filepath.Base(path)] = template
```
stevekuznetsov, line 58: This seems like something that we should have a test to catch

cmd/pj-rehearse/main.go
droslean: I changed the cleanup process to delete a collection of a configMaps that have the specific label.

pkg/config/template.go
droslean: WIP. I will prepare tests for every case.

cmd/pj-rehearse/main.go
droslean: We will do all the necessary cleanups in a `gracefullExit()` function. This can be done in a follow-up. For now, just deferring is ok.

petr-muller: I **really** think we should do the proper cleanup in this PR, not a followup.

cmd/pj-rehearse/main.go
```diff
@@ -168,14 +168,34 @@ func main() {
...
+	cmClient, err := rehearse.NewCMClient(clusterConfig, masterConfig.Prow.ProwJobNamespace, o.dryRun)
+	if err != nil {
+		logger.WithError(err).Error("could not create a configMap client")
+		gracefulExit(o.noFail, misconfigurationOutput)
+	}
+
+	cmController := config.NewTemplateCMController(cmClient, prNumber, logger, changedTemplates)
+	if err := cmController.CreateCMTemplates(); err != nil {
+		logger.WithError(err).Error("couldn't create template configMap")
+		gracefulExit(o.noFail, "")
```
petr-muller: Two things:

1. Do not pass an empty string as a message that's intended to be user-facing
2. This will still leave some CMs behind if `CreateCMTemplates` manage to create some and then fails.

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 func isYAML(file string, info os.FileInfo) bool {
 	return !info.IsDir() && (filepath.Ext(file) == ".yaml" || filepath.Ext(file) == ".yml")
 }
+
+// TemplateCMController holds the details needed for the configmap controller
+type TemplateCMController struct {
+	cmclient  corev1.ConfigMapInterface
+	prNumber  int
+	logger    *logrus.Entry
+	templates map[string]*templateapi.Template
```
petr-muller: Can we use the `CiTemplates` here?

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+}
+
+// NewTemplateCMController creates a new TemplateCMController
+func NewTemplateCMController(cmclient corev1.ConfigMapInterface, prNumber int, logger *logrus.Entry, templates map[string]*templateapi.Template) *TemplateCMController {
```
petr-muller: 1. Can we use the `CiTemplates` type instead of the raw map type?
2. I think the API would be more clear if `templates` was a parameter of `CreateCMTemplates` method instead.

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		cmName := GetTempCMName(c.prNumber, template.Name)
+		cm := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cmName,
+				Labels: map[string]string{
+					"ci.openshift.org/rehearse": "rehearse-temporary-cm",
```
petr-muller: 1. `ci.openshift.org/rehearse` should probably be a constant
2. I think the label needs to include a PR number, otherwise we have no way to search (and e.g. clean up) only the CMs created for a given PR

cmd/pj-rehearse/main.go
```diff
@@ -226,4 +243,16 @@ func main() {
 		gracefulExit(o.noFail, jobsFailureOutput)
 	}
 	logger.Info("All jobs were rehearsed successfully")
+
+	logger.Info("Cleanup temporary configMaps")
+	if err := cmCleanup(cmController); err != nil {
```
petr-muller: Placing the cleanup here will still only clean up the CMs when no problems occur. All the places between CM creation and here that call `gracefulExit` for different reasons will cause `pj-rehearse` to leak CMs on the cluster, which I believe we should not do.

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+				},
+			},
+			Data: map[string]string{filename: buf.String()},
+		}
+
+		c.logger.WithField("template-name", template.Name).Info("creating rehearsal configMap for template")
```
petr-muller: Should we also log the name of the actual temporary CM for this template?

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		if _, err := c.cmclient.Create(cm); err != nil && !kerrors.IsAlreadyExists(err) {
```
petr-muller: What is the reasoning behind the suppressing the `AlreadyExists` error? That does not seem right. If the CM exists, it may contain some stale old content from some previous run. That would be a problem, we need the rehearsals to use the content that came with the PR.

pkg/config/template.go
```diff
@@ -47,3 +62,68 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		}
+	}
+	return kutilerrors.NewAggregate(errors)
+}
+
+// CleanupCMTemplates deletes all the configMaps that have been created for the changed templates.
+func (c *TemplateCMController) CleanupCMTemplates() error {
+	c.logger.Info("deleting temporary template configMaps")
+	if err := c.cmclient.DeleteCollection(&metav1.DeleteOptions{}, metav1.ListOptions{
+		LabelSelector: fields.Set{"ci.openshift.org/rehearse": "rehearse-temporary-cm"}.AsSelector().String()}); err != nil {
```
petr-muller: This would delete even CMs created by rehearsals for a different PR (see https://github.com/openshift/ci-operator-prowgen/pull/100/files#diff-33dd159268780d36aa2b7bd9253f7a29R102)

pkg/rehearse/jobs.go
```diff
@@ -25,6 +31,8 @@ import (
 	"k8s.io/test-infra/prow/pjutil"
 
 	"github.com/openshift/ci-operator-prowgen/pkg/config"
+
+	"k8s.io/api/core/v1"
```
petr-muller: This third-party import should be somewhere higher, before the `ci-operator-prowgen`

pkg/rehearse/jobs.go
```diff
@@ -164,7 +187,7 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
 
 // ConfigureRehearsalJobs filters the jobs that should be rehearsed, then return a list of them re-configured with the
 // ci-operator's configuration inlined.
-func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool) []*prowconfig.Presubmit {
+func ConfigureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool, cmClient coreclientset.ConfigMapInterface, templates map[string]*templateapi.Template) []*prowconfig.Presubmit {
```
petr-muller: Please do not change the `config.Presubmits` type back to `map[string][]prowconfig.Presubmit`

pkg/rehearse/jobs.go
```diff
@@ -183,6 +206,14 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 				continue
 			}
 
+			exists, index, volumeName, err := hasChangedTemplateVolume(rehearsal.Spec.Volumes, cmClient, templates)
+			if err != nil {
+				jobLogger.WithError(err).Error("failed to check if job uses a changed template")
```
petr-muller: We should `continue` on error to avoid running a possibly invalid rehearsal.

pkg/rehearse/jobs.go
```diff
@@ -191,6 +222,34 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 	return rehearsals
 }
 
+// hasChangedTemplateVolume is checking if a list of volumes are using a template's configMap that has been changed.
+// It returns also the index of the volumes that has the changed template as well its name too.
+func hasChangedTemplateVolume(volumes []v1.Volume, cmClient coreclientset.ConfigMapInterface, templates map[string]*templateapi.Template) (bool, int, string, error) {
+	for index, volume := range volumes {
+		if volume.VolumeSource.ConfigMap != nil {
+			cm, err := cmClient.Get(volume.VolumeSource.ConfigMap.Name, metav1.GetOptions{})
```
petr-muller: I asked this earlier in https://github.com/openshift/ci-operator-prowgen/pull/100#discussion_r262529365 but there was no answer. Why do we need to touch the cluster here? We know which templates changed already, and we should be able to detect which jobs use them from the config. What do we need the cluster for?

The reason I am asking is that until now, `ConfigureRehearsalJobs` was able to run offline, entirely over the config data present in the repository. I'd prefer to keep it that way.

petr-muller (commented):
There is a bunch of smaller code comments and proposed logic fixes, and I continue to have concerns about these two areas:

1. We would be leaking a lot of CMs after we merge this.
2. I do not understand why `ConfigureRehearsalJobs` (via `hasChangedTemplateVolume`) needs to get data from the cluster.

pkg/config/template.go
droslean: Shall we remove it if alreadyExists first?

pkg/config/template.go
droslean: Yes I will update that.

pkg/rehearse/jobs.go
droslean: oh. sry lost in the rebase :)

droslean:
Back to WIP. I will remove the logic of getting the cm from the cluster to find which template file is using, since we have this information in [here](https://github.com/openshift/ci-operator-prowgen/blob/master/cmd/ci-operator-prowgen/main.go#L210). Thanks @petr-muller for mention it.

pkg/config/template.go
droslean: Maybe, if already exists, just update it? to avoid the re-creation

pkg/config/template.go
droslean: The changed templates won't change during the runtime. It is orthogonal to the controller to hold that information.

pkg/config/template.go
petr-muller: Updating the content sounds good, should not break anything.

pkg/config/template.go
petr-muller: It was just a suggestion. I was trying to think about the code from the API/module point of view.

stevekuznetsov:
This is getting fairly large, might be good to break it up if it needs to grow any more

cmd/pj-rehearse/main.go
```diff
@@ -87,9 +87,21 @@ an infrastructure issue.`
 
 pj-rehearse rehearsed jobs and at least one of them failed. This means that
 job would fail when executed against the current HEAD of the target branch.`
+
+	deleteConfigMapsFailureOutput = ` [ERROR] pj-rehearse: delete configmaps failure
+
+pj-rehearse could not delete the collection of the temporary configmaps that have
+been created for this execution. That means that they will still exist in the cluster
+and any administrator should be notified for manual integration.`
```
stevekuznetsov: Shouldn't we clean up at the start of the run or use a replacing call? "Contact an admin" doesn't scale

cmd/pj-rehearse/main.go
```diff
@@ -87,9 +87,21 @@ an infrastructure issue.`
...
 )
 
-func gracefulExit(suppressFailures bool, message string) {
+func gracefulExit(suppressFailures bool, message string, cmController *config.TemplateCMController) {
```
stevekuznetsov: Why are we not using the https://golang.org/pkg/runtime/#Goexit method?

cmd/pj-rehearse/main.go
```diff
@@ -140,26 +152,26 @@ func main() {
...
 	if masterConfig.Prow == nil || prConfig.Prow == nil {
 		logger.WithError(err).Error("could not load Prow configs from base or tested revision of release repo")
-		gracefulExit(o.noFail, misconfigurationOutput)
+		gracefulExit(o.noFail, misconfigurationOutput, nil)
 	}
 	// We always need PR versions of templates and ciop config, otherwise we cannot provide them to rehearsed jobs
 	if prConfig.Templates == nil || prConfig.CiOperator == nil {
 		logger.WithError(err).Error("could not load template/ci-operator configs from tested revision of release repo")
-		gracefulExit(o.noFail, misconfigurationOutput)
+		gracefulExit(o.noFail, misconfigurationOutput, nil)
```
stevekuznetsov: nit: if you find yourself making one call with a new argument and N with `nil`, make a new function:


```go
gracefulExitWithConfigMapCleanup(o.noFail, configmapOutput, controller)
```

(not that this is the best solution for this specific instance, but just as a nit

pkg/config/template.go
```diff
@@ -47,3 +64,74 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 func isYAML(file string, info os.FileInfo) bool {
 	return !info.IsDir() && (filepath.Ext(file) == ".yaml" || filepath.Ext(file) == ".yml")
 }
+
+// TemplateCMController holds the details needed for the configmap controller
+type TemplateCMController struct {
```
stevekuznetsov: `"controller"` has a specific meaning, and this is not a controller ;)

pkg/config/template.go
```diff
@@ -47,3 +64,74 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+	c.logger.Info("deleting temporary template configMaps")
+	if err := c.cmclient.DeleteCollection(&metav1.DeleteOptions{},
+		metav1.ListOptions{LabelSelector: fields.Set{rehearseLabel: fmt.Sprintf("rehearse-temporary-cm-%d", c.prNumber)}.AsSelector().String()}); err != nil {
+		return err
+	}
+	return nil
+}
+
+// GetTempCMName converts a template name to a temporary name including the pr number.
+func GetTempCMName(prNumber int, templateName string) string {
```
stevekuznetsov: if we hash the data that is used to fill in the configmap, we will ensure we never have collisions and cleanup can be best-effort or even out-of-band, that seems more robust than this approach

pkg/config/template.go
```diff
@@ -47,3 +64,74 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		cmName := GetTempCMName(c.prNumber, template.Name)
+		cm := &v1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: cmName,
+				Labels: map[string]string{
+					rehearseLabel: fmt.Sprintf("rehearse-temporary-cm-%d", c.prNumber),
```
stevekuznetsov: I would suggest re-using similar names to the upstream ones:

```yaml
  labels:
    created-by-prow: "true"
    prow.k8s.io/id: ff1f5cb0-4343-11e9-b256-0a58ac10d9a5
    prow.k8s.io/job: pull-ci-openshift-openshift-azure-master-e2e-azure-forceupdate
    prow.k8s.io/refs.org: openshift
    prow.k8s.io/refs.pull: "1261"
    prow.k8s.io/refs.repo: openshift-azure
    prow.k8s.io/type: presubmit
```

pkg/config/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"k8s.io/client-go/kubernetes/typed/core/v1/fake"
+	coretesting "k8s.io/client-go/testing"
+)
+
+const templatesPath = "../../test/pj-rehearse-integration/master/ci-operator/templates"
+
+func TestGetTemplates(t *testing.T) {
+	expectCiTemplates := getBaseCiTemplates(t)
+	if templates, err := getTemplates(templatesPath); err != nil {
+		t.Fatal(err)
```
stevekuznetsov: This err and the following one don't have enough context for someone who accidentally breaks the test

pkg/config/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+	} else if !equality.Semantic.DeepEqual(templates, expectCiTemplates) {
+		t.Fatalf("Diff found %s", diff.ObjectReflectDiff(expectCiTemplates, templates))
+	}
+}
+
+var (
```
stevekuznetsov: why are these globals?

pkg/config/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+			t.Fatalf("Configmap name:\nExpected: %s\nFound: %s", expectedCmName, cm.ObjectMeta.Name)
+		}
+
+		if !reflect.DeepEqual(cm.ObjectMeta.Labels, expectedCmLabels) {
+			t.Fatalf("Configmap labels\nExpected: %#v\nFound: %#v", expectedCmLabels, cm.ObjectMeta.Labels)
+		}
+
+		return true, nil, nil
+	})
+	cs.Fake.AddReactor("*", "configmaps", func(action coretesting.Action) (handled bool, ret runtime.Object, err error) {
```
stevekuznetsov: you bind this to all actions, not just delete, so it will panic if someone changes the code under test

pkg/rehearse/jobs.go
```diff
@@ -10,13 +10,19 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/api/core/v1"
```
stevekuznetsov, line 13: nit: imports in this file should be split into correct blocks

pkg/rehearse/jobs.go
```diff
@@ -53,6 +59,21 @@ func NewProwJobClient(clusterConfig *rest.Config, namespace string, dry bool) (p
 	return pjcset.ProwV1().ProwJobs(namespace), nil
 }
 
+// NewCMClient creates a configMap client with a dry run capability
+func NewCMClient(clusterConfig *rest.Config, namespace string, dry bool) (coreclientset.ConfigMapInterface, error) {
+	if dry {
+		c := &fake.FakeCoreV1{Fake: &coretesting.Fake{}}
```
stevekuznetsov: this is neat

pkg/rehearse/jobs_test.go
```diff
@@ -754,3 +756,221 @@ func makeBasePresubmit() *prowconfig.Presubmit {
...
+				{
+					Name:      "job-definition",
+					MountPath: "/tmp/test",
+					SubPath:   "test-template.yaml",
+				},
+			},
+			jobVolumes: []v1.Volume{
+				{
+					Name: "job-definition",
+					VolumeSource: v1.VolumeSource{
```
stevekuznetsov: you could make a helper  `func volumeSource(name, key string) v1.VolumeSource` to make these tests more readable

pkg/rehearse/jobs.go
```diff
@@ -183,6 +204,11 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 				continue
 			}
 
+			exists, index, templateName := hasChangedTemplateVolume(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates)
+			if exists {
+				rehearsal.Spec.Volumes[index].VolumeSource.ConfigMap.Name = config.GetTempCMName(prNumber, templateName)
```
stevekuznetsov: How do we test this?

cmd/pj-rehearse/main.go
droslean: This is our last line of defense!! If that fails, and the pj-rehearse tool will exit, how we will manage to delete those configmaps? Do you think that its a better idea to add some `ttl` label and update the ttl controller to clean up also configmaps?

pkg/config/template.go
droslean: In the beginning, I wanted to make it as a controller, and I stacked with the name :). I will rename that.

pkg/config/template.go
droslean: Hashing the data seems the best solution since recently I discovered that if someone run the tool with `--local` it will always create a cm with the same name.  But still, hashing the same data, with end up in a collision.

pkg/config/template_test.go
droslean: I wasn't able to find the correct verb for `DeleteCollection`. I will take a further look at it.

pkg/rehearse/jobs.go
droslean: `TestConfigureRehearsalJobs` should be updated. This PR is already getting bigger and bigger. I would like to do this in a follow-up

cmd/pj-rehearse/main.go
droslean: @stevekuznetsov @petr-muller any thoughts on that?

pkg/config/template.go
droslean: Something like
```yaml
labels:
  created-by-pj-rehearse: "true"
  ci.openshift.org/rehearse/pull: "1234"
```

cmd/pj-rehearse/main.go
stevekuznetsov: Hash the input files to the maps, so it doesn't ever collide, and do an out-of-band cleanup IMO

pkg/config/template.go
stevekuznetsov: If you hash and collide, don't bother with anything -- a collision on the hash of the contents means that the CM in the cluster already has the right data in it

pkg/config/template_test.go
stevekuznetsov: @deads2k do you know?

pkg/config/template.go
droslean: that it a good approach. I will make the changes in this PR

pkg/config/template_test.go
droslean: "deletecollection" verb doesn't work. Bug in upstream? I will investigate

droslean: Ok I found it. `delete-collection` :)

droslean:
retest

cmd/pj-rehearse/main.go
droslean: I am not afraid of a collision.  If the delete will fail for a reason that it's related to the cluster, the cm will stay there forever.

petr-muller: I think a failure to clean up temporary CMs should not even be a user facing failure and does not need such verbose message. If we fail to clean up something after everything else went well, just log it and finish.

pkg/config/template.go
```diff
@@ -47,3 +70,105 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+// CreateCMTemplates creates configMaps for all the changed templates.
+func (c *TemplateCMManager) CreateCMTemplates() error {
+	var errors []error
+	for filename, template := range c.templates {
+		templateData, err := GetTemplateData(template)
+		if err != nil {
+			errors = append(errors, err)
+		}
+
+		cmName := GetTempCMName(template.Name, []string{filename, templateData})
```
petr-muller: I would prefer the `filename` and `templateData` to be proper arguments instead of the API requiring a slice (why should I care as a caller) but whatever.

pkg/config/template.go
```diff
@@ -47,3 +70,105 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+					rehearseLabelPull: strconv.Itoa(c.prNumber),
+				},
+			},
+			Data: map[string]string{filename: templateData},
+		}
+
+		c.logger.WithFields(logrus.Fields{"template-name": template.Name, "cm-name": cmName}).Info("creating rehearsal configMap for template")
+		if _, err := c.cmclient.Create(cm); err != nil {
+			if kerrors.IsAlreadyExists(err) {
+				if _, err := c.cmclient.Update(cm); err != nil {
```
petr-muller, line 116: In theory we do not need to update, because we hash the content to the name, so the content is guaranteed to be identical, but I'd like to keep this - it makes the logic robust to the changes we could do to the names.

pkg/config/template.go
```diff
@@ -47,3 +70,105 @@ func getTemplates(templatePath string) (CiTemplates, error) {
...
+			createByRehearse:  "true",
+			rehearseLabelPull: strconv.Itoa(c.prNumber),
+		}.AsSelector().String()}); err != nil {
+		return err
+	}
+	return nil
+}
+
+func GetTempCMName(templateName string, inputs []string) string {
+	return fmt.Sprintf("rehearse-%s-%s", inputHash(inputs), templateName)
```
petr-muller, line 142: Hm, how the long will this be? How long will be the hash? Won't we hit some limits?

pkg/rehearse/jobs.go
```diff
@@ -183,6 +203,13 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 				continue
 			}
 
+			exists, index, templateKey := hasChangedTemplateVolume(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates)
+			if exists {
+				if templateData, err := config.GetTemplateData(templates[templateKey]); err == nil {
+					rehearsal.Spec.Volumes[index].VolumeSource.ConfigMap.Name = config.GetTempCMName(templates[templateKey].Name, []string{templateKey, templateData})
+				}
```
petr-muller, line 213: If there **was** an error, we should `continue` to avoid submitting possibly broken rehearsal.  If the job uses some changed template but we fail to get its data, there's something wrong.

pkg/rehearse/jobs.go
```diff
@@ -191,6 +218,26 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+	for _, volumeMount := range volumeMounts {
+		if _, ok := templates[volumeMount.SubPath]; ok {
+			templateKey = volumeMount.SubPath
+			volumeName = volumeMount.Name
+		}
+	}
+
+	for index, volume := range volumes {
+		if volume.Name == volumeName {
+			return true, index, templateKey
```
petr-muller, line 237: So this looks for a first occurrence... Is it sufficient? Won't we need to fixup multiple references, maybe to a cluster profile **and** the template?

petr-muller (commented):
Getting there... Just few small things and some questions.

pkg/config/template.go
droslean: The hash is always 5 characters. `rehearse` is 8 and the biggest cm name that we currently have is 45. So in total, we are in 60 characters maximum (with the dashes)

pkg/rehearse/jobs.go
droslean: We don't need to change the secrets. Only the template.

pkg/rehearse/jobs.go
droslean: addressed

pkg/config/template.go
droslean: addressed

cmd/pj-rehearse/main.go
droslean: addressed

pkg/rehearse/jobs.go
petr-muller: OK

petr-muller (approved):
/lgtm

I think this one is good to go.

PR #101, opened 2019-03-06
Rehearse jobs that use changed ci-operator configs

petr-muller:
Post-https://github.com/openshift/ci-operator-prowgen/pull/96, we are able to detect which ci-operator configs changed in a PR. To start rehearsing the jobs affected by changed ci-operator configs, we needed to implement these two remaining capabilities:

1. Walk through all presubmits and find which ones reference one of the changed configs
2. Merge the two discovered sets of rehearsal jobs without creating duplicates.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -163,11 +163,9 @@ func main() {
 	}
 
 	// We can only detect changes if we managed to load both ci-operator config versions
+	changedCiopConfigs := config.CompoundCiopConfig{}
 	if masterConfig.CiOperator != nil && prConfig.CiOperator != nil {
-		changedCiopConfigs := diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
-		for name := range changedCiopConfigs {
-			logger.WithField("ciop-config", name).Info("Changed ci-operator config")
-		}
+		changedCiopConfigs = diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
 	}
 
 	// We can only detect changes if we managed to load both CI template versions
@@ -201,8 +199,10 @@ func main() {
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
-	changedPresubmits := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
-	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, prConfig.CiOperator, prNumber, loggers, o.allowVolumes)
+	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
+	toRehearse.AddAll(diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger))
+
+	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		os.Exit(0)
--- a/cmd/pj-rehearse/main_test.go
+++ b/cmd/pj-rehearse/main_test.go
@@ -131,6 +131,7 @@ func getRehersalsHelper(logger *logrus.Entry, prNumber int) ([]*prowconfig.Presu
 	candidateConfigPath := filepath.Join(candidatePath, config.ConfigInRepoPath)
 	candidateJobConfigPath := filepath.Join(candidatePath, config.JobConfigInRepoPath)
 	candidateCiopConfigPath := filepath.Join(candidatePath, config.CiopConfigInRepoPath)
+	masterCiopConfigPath := filepath.Join(masterPath, config.CiopConfigInRepoPath)
 	masterConfigPath := filepath.Join(masterPath, config.ConfigInRepoPath)
 	masterJobConfigPath := filepath.Join(masterPath, config.JobConfigInRepoPath)
 
@@ -146,11 +147,18 @@ func getRehersalsHelper(logger *logrus.Entry, prNumber int) ([]*prowconfig.Presu
 	if err != nil {
 		return nil, fmt.Errorf("failed to load ci-operator config: %v", err)
 	}
+	ciopMasterConfig, err := config.CompoundLoad(masterCiopConfigPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load ci-operator config: %v", err)
+	}
 
 	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
 	if len(changedPresubmits) == 0 {
 		return nil, fmt.Errorf("Empty changedPresubmits was not expected")
 	}
+
+	changedCiopConfigs := diffs.GetChangedCiopConfigs(ciopMasterConfig, ciopPrConfig, logger)
+	changedPresubmits.AddAll(diffs.GetPresubmitsForCiopConfigs(prowPRConfig, changedCiopConfigs, logger))
 
 	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, ciopPrConfig, prNumber, rehearse.Loggers{Job: logger, Debug: logger}, false)
 	return rehearsals, nil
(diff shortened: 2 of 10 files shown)


cmd/pj-rehearse/main_test.go
```diff
@@ -146,12 +147,19 @@ func getRehersalsHelper(logger *logrus.Entry, prNumber int) ([]*prowconfig.Presu
...
+		return nil, fmt.Errorf("failed to load ci-operator config: %v", err)
+	}
 
 	changedPresubmits := diffs.GetChangedPresubmits(prowConfig, prowPRConfig, logger)
 	if len(changedPresubmits) == 0 {
 		return nil, fmt.Errorf("Empty changedPresubmits was not expected")
 	}
 
+	changedCiopConfigs := diffs.GetChangedCiopConfigs(ciopMasterConfig, ciopPrConfig, logger)
+	changedPresubmits.AddAll(diffs.GetPresubmitsForCiopConfigs(prowPRConfig, changedCiopConfigs, logger))
```
stevekuznetsov, line 161: The fact that you needed to do this in the test, too, is smelly

pkg/config/jobs_test.go
```diff
@@ -0,0 +1,118 @@
...
+			prowconfig.Presubmit{JobBase: prowconfig.JobBase{Name: "source-job-for-org-source-repo"}},
+		}},
+		destination: Presubmits{"org/destination-repo": {
+			prowconfig.Presubmit{JobBase: prowconfig.JobBase{Name: "destination-job-for-org-destination-repo"}}}},
+		expected: Presubmits{
+			"org/source-repo":      {prowconfig.Presubmit{JobBase: prowconfig.JobBase{Name: "source-job-for-org-source-repo"}}},
+			"org/destination-repo": {prowconfig.Presubmit{JobBase: prowconfig.JobBase{Name: "destination-job-for-org-destination-repo"}}},
+		},
+	}, {
+		description: "merge jobs with same name for a single repo, result has the one originally in destination",
```
stevekuznetsov, line 60: Since we need unique names to work this might be better as an error, but meh

pkg/config/release.go
```diff
@@ -20,6 +20,8 @@ const (
 	CiopConfigInRepoPath = "ci-operator/config"
 	// TemplatesPath is the path of the templates from release repo
 	TemplatesPath = "ci-operator/templates"
+	// Name of the configmap that stores all ci-operator configs
+	CiOperatorConfigsCMName = "ci-operator-configs"
```
stevekuznetsov, line 26: This is not true -- we are sharded. See the `ConfigMapName()` method on the `config.Info` struct

pkg/rehearse/jobs.go
```diff
@@ -166,7 +164,7 @@ func inlineCiOpConfig(job *prowconfig.Presubmit, targetRepo string, ciopConfigs
 
 // ConfigureRehearsalJobs filters the jobs that should be rehearsed, then return a list of them re-configured with the
 // ci-operator's configuration inlined.
-func ConfigureRehearsalJobs(toBeRehearsed map[string][]prowconfig.Presubmit, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool) []*prowconfig.Presubmit {
+func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool) []*prowconfig.Presubmit {
```
stevekuznetsov, line 167: Might have been nicer to do the type changes in a separate PR

cmd/pj-rehearse/main_test.go
petr-muller: True, that test basically mimics our main().

pkg/config/jobs_test.go
petr-muller: It cannot be an error. In rehearse, when both ci-operator config and a presubmit that uses it change, we'll want to rehearse the same job for two reasons, so it ends up being returned by two different "what jobs do we need to rehearse" methods, and those results need to be merged.

pkg/config/release.go
petr-muller: I'm aware of that, I want to tackle that tomorrow. IMHO it's worth a separate PR. I only moved this constant, not created it. Rehearse is a little broken after the sharding anyways.

pkg/rehearse/jobs.go
petr-muller: Maybe. It's basically just a rename, I did a separate commit, and it was mostly driven by the need for the `AddAll` method.

pkg/diffs/diffs.go
```diff
@@ -141,3 +141,33 @@ func GetChangedTemplates(masterTemplates, prTemplates map[string]*templateapi.Te
...
+
+func GetPresubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfigs config.CompoundCiopConfig, logger *logrus.Entry) config.Presubmits {
+	ret := config.Presubmits{}
+
+	for repo, jobs := range prowConfig.JobConfig.Presubmits {
+		for _, job := range jobs {
+			if job.Agent != string(pjapi.KubernetesAgent) {
+				continue
+			}
+			for _, container := range job.Spec.Containers {
```
droslean: there is no need to iterate here. just use `job.Spec.Containers[0]`
There is a validation already happening in the upstream library.

cmd/pj-rehearse/main.go
```diff
@@ -188,8 +186,10 @@ func main() {
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
-	changedPresubmits := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
-	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, prConfig.CiOperator, prNumber, loggers, o.allowVolumes)
+	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
+	toRehearse.AddAll(diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger))
```
droslean, line 203: So, the intention here is to add all the jobs that are using the changed ci-op config?

droslean (commented):
In general, this PR is ok.  But still there are some cases missing.

If there is a changed ci-op config and a changed presubmit that its using that config, there is no need to do all of this, and instead we can just rehearse the changed presubmit with the changed ci-op config inlined. I am pretty sure that in most cases this is what will happen since if someone will change his ci-op config, prowgen will generate different presubmit. 

So, the only thing left in this PR is to do this procedure only if there is no other changed presubmit that uses the changed ci-op config.

pkg/diffs/diffs.go
petr-muller: WIll do.

cmd/pj-rehearse/main.go
petr-muller: All the jobs that are using **some** of the changed ci-op configs, there could be more...

`AddAll` prevents job being present twice, so if a job was already in `toRehearse` because it was changed directly, it's not added again.

petr-muller:
> If there is a changed ci-op config and a changed presubmit that its using that config, there is no need to do all of this, and instead we can just rehearse the changed presubmit with the changed ci-op config inlined. I am pretty sure that in most cases this is what will happen since if someone will change his ci-op config, prowgen will generate different presubmit.
> 
> So, the only thing left in this PR is to do this procedure only if there is no other changed presubmit that uses the changed ci-op config.

Can you elaborate on the changes that you are requesting?  We detect which jobs changed directly. Then we detect which ciop configs changed and which jobs are using them, and merge these jobs with those changed directly. The resulting set of jobs will be rehearsed. What do you think should be done differently?

droslean:
> Can you elaborate on the changes that you are requesting? We detect which jobs changed directly. Then we detect which ciop configs changed and which jobs are using them, and merge these jobs with those changed directly. The resulting set of jobs will be rehearsed. What do you think should be done differently?

I wanted to be more selective in which jobs we should rehearse instead of running them all when we detect a changed ci-op config. I think what its implemented in this PR is a good start and we can be more selective in the future.

petr-muller:
Addressed the actionable comments.

pkg/config/release.go
petr-muller: Addressed in https://github.com/openshift/ci-operator-prowgen/pull/104

pkg/config/jobs_test.go
stevekuznetsov: Ah

stevekuznetsov (approved):
/lgtm

PR #102, opened 2019-03-07
The new upgrade job is supported as part of the ci-operator

smarterclayton:
Instead of mounting the template, the ci-operator can run the job itself.
Use the upgrade=true flag only to test this logic while we verify that
upgrade tests (and their attendant complexity) don't destabilize other
jobs.

Examples of the job generated by this PR are in openshift/release#3030.  Will be testing openshift/ci-operator#288 before we merge


smarterclayton:
@droslean

droslean:
@smarterclayton That seems to caused by bumping the ci-operator libraries,
Can you remove those `canonical_go_repository: ""` from `test/pj-rehearse-integration/expected_jobs.yaml` in this PR?

smarterclayton:
Oh, I thought this was rehearsing against the real repositories.  Duh, fixing.

smarterclayton:
Fixed

PR #108, opened 2019-03-12
rehearse: Add an option to dump metrics after execution

petr-muller:
My goal is to dump an artifact with a complete record of what happened - what changes we detected, what jobs we decided to rehearse, why, and what the result was. This artifact could then be scraped from all runs and analysed, and later uploaded to some document store.

WDYT?

Currently, the JSON dump looks like this (comments added by me):
```json
{
  "org": "openshift",
  "repo": "release",
  "pr": 3116,
  "changed_ciop_configs": [
    "openshift-cluster-api-master.yaml"
  ],
  "changed_presubmits": [
    "pull-ci-openshift-cluster-api-master-git-history"
  ],
  "changed_templates": [],
  "opportunities": {
    "pull-ci-openshift-cluster-api-master-git-history": [
      "direct-change",
      "ci-operator-config-change"
    ],
    "pull-ci-openshift-cluster-api-master-gofmt": [
      "ci-operator-config-change"
    ],
    "pull-ci-openshift-cluster-api-master-govet": [
      "ci-operator-config-change"
    ],
    "pull-ci-openshift-cluster-api-master-images": [
      "ci-operator-config-change"
    ],
    "pull-ci-openshift-cluster-api-master-unit": [
      "ci-operator-config-change"
    ]
  },
  "actual": [
    "rehearse-3116-pull-ci-openshift-cluster-api-master-git-history",
    "rehearse-3116-pull-ci-openshift-cluster-api-master-gofmt",
    "rehearse-3116-pull-ci-openshift-cluster-api-master-govet",
    "rehearse-3116-pull-ci-openshift-cluster-api-master-images",
    "rehearse-3116-pull-ci-openshift-cluster-api-master-unit"
  ],
  "execution": {
    "submitted": [
      "rehearse-3116-pull-ci-openshift-cluster-api-master-git-history",
      "rehearse-3116-pull-ci-openshift-cluster-api-master-gofmt",
      "rehearse-3116-pull-ci-openshift-cluster-api-master-govet",
      "rehearse-3116-pull-ci-openshift-cluster-api-master-images",
      "rehearse-3116-pull-ci-openshift-cluster-api-master-unit"
    ],
    "failed": null,
    "successful": null
  }
}
```

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -44,6 +44,7 @@ type options struct {
 	local        bool
 	allowVolumes bool
 	debugLogPath string
+	metricsPath  string
 
 	releaseRepoPath string
 	rehearsalLimit  int
@@ -60,6 +61,7 @@ func gatherOptions() options {
 
 	fs.StringVar(&o.debugLogPath, "debug-log", "", "Alternate file for debug output, defaults to stderr")
 	fs.StringVar(&o.releaseRepoPath, "candidate-path", "", "Path to a openshift/release working copy with a revision to be tested")
+	fs.StringVar(&o.metricsPath, "metrics-output", "", "Path to a file where JSON metrics will be dumped after rehearsal")
 
 	fs.IntVar(&o.rehearsalLimit, "rehearsal-limit", 15, "Upper limit of jobs attempted to rehearse (if more jobs would be rehearsed, none will)")
 
@@ -113,6 +115,9 @@ func rehearseMain() int {
 		return gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
+	metrics := rehearse.NewMetrics(o.metricsPath)
+	defer metrics.Dump()
+
 	var jobSpec *pjdwapi.JobSpec
 	if o.local {
 		if jobSpec, err = config.NewLocalJobSpec(o.releaseRepoPath); err != nil {
@@ -127,11 +132,13 @@ func rehearseMain() int {
 	}
 
 	prFields := logrus.Fields{prowgithub.OrgLogField: jobSpec.Refs.Org, prowgithub.RepoLogField: jobSpec.Refs.Repo}
+	metrics.Org = jobSpec.Refs.Org
+	metrics.Repo = jobSpec.Refs.Repo
 	logger := logrus.WithFields(prFields)
 
 	if jobSpec.Type != pjapi.PresubmitJob {
 		logger.Info("Not able to rehearse jobs when not run in the context of a presubmit job")
-		// Exiting successfuly will make pj-rehearsal job not fail when run as a
+		// Exiting successfully will make pj-rehearsal job not fail when run as a
 		// in a batch job. Such failures would be confusing and unactionable
 		return 0
 	}
@@ -140,6 +147,7 @@ func rehearseMain() int {
 	if o.local {
 		prNumber = int(time.Now().Unix())
 	}
+	metrics.Pr = prNumber
 	logger = logrus.WithField(prowgithub.PrLogField, prNumber)
 	logger.Info("Rehearsing Prow jobs for a configuration PR")
 
@@ -174,12 +182,14 @@ func rehearseMain() int {
 	changedCiopConfigs := config.CompoundCiopConfig{}
 	if masterConfig.CiOperator != nil && prConfig.CiOperator != nil {
 		changedCiopConfigs = diffs.GetChangedCiopConfigs(masterConfig.CiOperator, prConfig.CiOperator, logger)
+		metrics.RecordChangedCiopConfigs(changedCiopConfigs)
 	}
 
 	changedTemplates := make(config.CiTemplates)
 	// We can only detect changes if we managed to load both CI template versions
 	if masterConfig.Templates != nil && prConfig.Templates != nil {
 		changedTemplates = diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
+		metrics.RecordChangedTemplates(changedTemplates)
 	}
 
 	namespace := prConfig.Prow.ProwJobNamespace
@@ -224,9 +234,15 @@ func rehearseMain() int {
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
 	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
-	toRehearse.AddAll(diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger))
+	metrics.RecordChangedPresubmits(toRehearse)
+	metrics.RecordOpportunity(toRehearse, "direct-change")
+
+	presubmitsWitchChangedCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger)
+	metrics.RecordOpportunity(presubmitsWitchChangedCiopConfigs, "ci-operator-config-change")
+	toRehearse.AddAll(presubmitsWitchChangedCiopConfigs)
 
 	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates)
+	metrics.RecordActual(rehearsals)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		return 0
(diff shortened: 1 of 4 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -89,7 +89,7 @@ pj-rehearse rehearsed jobs and at least one of them failed. This means that
 job would fail when executed against the current HEAD of the target branch.`
 )
 
-func gracefulExit(suppressFailures bool, message string) {
+func gracefulExit(suppressFailures bool, message string, metrics *rehearse.Metrics) {
```
stevekuznetsov: Same comments as to Nikos in the other PR:

1. for this case, let's use https://golang.org/pkg/runtime/#Goexit and defer the `os.Exit()`
2. passing `nil` around without telling people what it is at the callsite is confusing -- would have been nicer to make a new method that accepts the metrics and keep the current one the same

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/openshift/ci-operator-prowgen/pkg/config"
+	"github.com/sirupsen/logrus"
+	"io/ioutil"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/apis/prowjobs/v1"
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+type ExecutionMetrics struct {
+	SubmittedRehearsals sets.String
```
stevekuznetsov: Might seralize better if we make these slices on output

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	FailedRehearsals    sets.String
+	PassedRehearsals    sets.String
+}
+
+func (m *ExecutionMetrics) RecordFailedRehearsal(job *v1.ProwJob) {
+	if m.FailedRehearsals == nil {
```
stevekuznetsov: usually prefer to instantiate a working object /shrug

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	prowconfig "k8s.io/test-infra/prow/config"
+)
+
+type ExecutionMetrics struct {
+	SubmittedRehearsals sets.String
+	FailedRehearsals    sets.String
+	PassedRehearsals    sets.String
+}
+
+func (m *ExecutionMetrics) RecordFailedRehearsal(job *v1.ProwJob) {
```
stevekuznetsov: If the fields are exported (as they need to be to serialize) we don't really need the helpers

stevekuznetsov (commented):
Looks sweet -- if we put this in a known place in GCS we can grab all of them with a batch operation really simply

cmd/pj-rehearse/main.go
petr-muller: Yeah this is the half-assed part - I knew Nikos is touching cleanups so I haven't tried being smart, and I will just adjust to whatever is the outcome in his PR.

pkg/rehearse/metrics.go
petr-muller: Yeah, probably nicer

pkg/rehearse/metrics.go
petr-muller: I like the readability at the callsite though - I don't like our `main()` cluttered with modifying the internal state of some random struct.

stevekuznetsov: How is it less readable to add to a map as a field?

cmd/pj-rehearse/main.go
```diff
@@ -194,36 +198,45 @@ func main() {
...
 			logger.WithError(err).Error("could not open debug log file")
-			gracefulExit(o.noFail, "")
+			gracefulExit(o.noFail, "", nil)
 		}
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
 	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
-	toRehearse.AddAll(diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger))
+	metrics.RecordPresubmits(toRehearse)
```
droslean: Since the time when you want to record the metrics doesn't matter, why not do the `metrics.Record` once, when we have already loaded the `ciopconfig` `templates` `presubmits`

pkg/rehearse/jobs.go
```diff
@@ -289,9 +293,11 @@ func (e *Executor) waitForJobs(jobs sets.String, selector string) (bool, error)
 			switch pj.Status.State {
 			case pjapi.FailureState, pjapi.AbortedState, pjapi.ErrorState:
 				e.loggers.Job.WithFields(fields).Error("Job failed")
+				e.Metrics.RecordFailedRehearsal(pj)
```
droslean: We have to change the logic here and avoid passing an `ExecutionMetrics`.  What are you doing here is logical and correct but the whole `waitForJobs` logic is not so good. (that's why I tried to do that refactor and pass the result in the channels here). Of course, we have to discuss this in one of our meetings.

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	if m.FailedRehearsals == nil {
+		m.FailedRehearsals = sets.NewString()
+	}
+	m.FailedRehearsals.Insert(job.Spec.Job)
+}
+
+func (m *ExecutionMetrics) RecordPassedRehearsal(job *v1.ProwJob) {
+	if m.PassedRehearsals == nil {
+		m.PassedRehearsals = sets.NewString()
```
droslean: I would prefer doing that inside a constructor, and void doing that check in the methods.

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	m.PassedRehearsals.Insert(job.Spec.Job)
+}
+
+func (m *ExecutionMetrics) RecordSubmittedRehearsal(job *v1.ProwJob) {
+	if m.SubmittedRehearsals == nil {
+		m.SubmittedRehearsals = sets.NewString()
```
droslean: same here

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,131 @@
...
+
+type ChangeMetrics struct {
+	CiopConfigs sets.String
+	Templates   sets.String
+	Presubmits  sets.String
+}
+
+func (m *ChangeMetrics) RecordCiopConfigs(configs config.CompoundCiopConfig) {
+	if m.CiopConfigs == nil {
+		m.CiopConfigs = sets.NewString()
```
droslean: same here, and the rest (you've got my point) :)

pkg/rehearse/metrics.go
petr-muller: Aye

cmd/pj-rehearse/main.go
petr-muller: Hm, not sure I understand what to do differently... This line records the presubmits that were changed (analogically to the `metrics.RecordTemplates` and `metrics.RecordCiopConfigs`) calls above, I do not see any better place to capture that number...

petr-muller:
I submitted a more polished version of this. Writing out the metrics is now behind an option that also allows to pass a path to the file where the content will be written. Writing the metrics is now done in a `defer`. The reviews were addressed. The output was slightly changed - the biggest change is probably that opportunities are now a `map[job-name][]reason` instead of vice versa.

@stevekuznetsov @droslean PTAL

pkg/rehearse/metrics.go
```diff
@@ -0,0 +1,101 @@
...
+	file   string
+}
+
+func NewMetrics(file string) *Metrics {
+	return &Metrics{
+		ChangedCiopConfigs: []string{},
+		ChangedPresubmits:  []string{},
+		ChangedTemplates:   []string{},
+
+		Opportunities: map[string][]string{},
```
droslean, line 44: lets `make()` this

petr-muller: Why? It seems that most of our code is using literals...

cmd/pj-rehearse/main.go
```diff
@@ -113,6 +115,9 @@ func rehearseMain() int {
 		return gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
+	metrics := rehearse.NewMetrics(o.metricsPath)
+	defer metrics.Dump()
```
droslean, line 119: if the `metricsPath` will be empty, we shouldn't gather metrics at all. right?

petr-muller: Correct, `Dump` is a NOP if the path is empty, so it's safe to defer like this.

https://github.com/openshift/ci-operator-prowgen/pull/108/files#diff-66bd766f197dd6b959d77192fcde3dbeR90

PR #110, opened 2019-03-14
rehearse: simplify termination & cleanup

petr-muller:
So I was trying to address https://github.com/openshift/ci-operator-prowgen/pull/108#discussion_r264877635 but haven't really found an obvious way how to use https://golang.org/pkg/runtime/#Goexit to make the termination a bit more straight and allow using `defer` for cleanup, but I saw this pattern sowewhere and actually like it.

@stevekuznetsov a pointer on an example of the `Goexit`-using termination would be helpful if you think that would still be better than this.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -87,34 +87,29 @@ an infrastructure issue.`
 
 pj-rehearse rehearsed jobs and at least one of them failed. This means that
 job would fail when executed against the current HEAD of the target branch.`
+	failedSetupOutput = `[ERROR] pj-rehearse: setup failure
+
+pj-rehearse failed to finish all setup necessary to perform job rehearsals.
+This is either a pj-rehearse bug or an infrastructure failure.`
 )
 
-func gracefulExit(suppressFailures bool, message string) {
+func gracefulExit(suppressFailures bool, message string) int {
 	if message != "" {
 		fmt.Fprintln(os.Stderr, message)
 	}
 
 	if suppressFailures {
-		os.Exit(0)
+		return 0
 	}
-	os.Exit(1)
+	return 1
 }
 
-func gracefulExitWithConfigMapCleanup(suppressFailures bool, message string, cmManager *config.TemplateCMManager) {
-	if cmManager != nil {
-		if err := cmManager.CleanupCMTemplates(); err != nil {
-			fmt.Fprintln(os.Stderr, "Failed to cleanup temporary configmaps")
-		}
-	}
-	gracefulExit(suppressFailures, message)
-}
-
-func main() {
+func rehearseMain() int {
 	o := gatherOptions()
 	err := validateOptions(o)
 	if err != nil {
 		logrus.WithError(err).Fatal("invalid options")
-		gracefulExit(o.noFail, misconfigurationOutput)
+		return gracefulExit(o.noFail, misconfigurationOutput)
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -87,34 +87,29 @@ an infrastructure issue.`
...
 
 	if suppressFailures {
-		os.Exit(0)
+		return 0
 	}
-	os.Exit(1)
+	return 1
 }
 
-func gracefulExitWithConfigMapCleanup(suppressFailures bool, message string, cmManager *config.TemplateCMManager) {
```
stevekuznetsov, line 103: Wait did we just nuke this?

cmd/pj-rehearse/main.go
```diff
@@ -228,29 +228,30 @@ func main() {
...
 	}
 	logger.Info("All jobs were rehearsed successfully")
+	return 0
+}
 
-	if err := cmManager.CleanupCMTemplates(); err != nil {
-		logger.WithError(err).Error("Failed to cleanup temporary configmaps")
-	}
+func main() {
+	os.Exit(rehearseMain())
```
stevekuznetsov, line 256: If we're doing this we can just `defer` the cleanup call at the top of `rehearseMain()` and not need to do the return value stuff.

stevekuznetsov (commented):
I was thinking to `defer` an `os.Exit()` call in `main()` and then instead call `runtime.Goexit()` where we called `os.Exit()` today to honor deferred statements.

cmd/pj-rehearse/main.go
petr-muller: Yeah, it's no longer needed, this way you can `defer` the cleanup (https://github.com/openshift/ci-operator-prowgen/pull/110/files#diff-06579b6b828bbff5df82617adb17e7bfR196)

cmd/pj-rehearse/main.go
petr-muller: How would we distinguish the success (`os.Exit(0)`) from failure (`os.Exit(1)`) of the whole rehearsal?

petr-muller:
> I was thinking to defer an os.Exit() call in main() and then instead call runtime.Goexit() where we called os.Exit() today to honor deferred statements.

But how would we communicate the `Exit` param (the desired return code) to the deferred `Exit` call?  I can imagine ways to do so but I haven't found anything elegant.

cmd/pj-rehearse/main.go
```diff
@@ -194,19 +189,24 @@ func main() {
...
 	cmManager := config.NewTemplateCMManager(cmClient, prNumber, logger, changedTemplates)
+	defer func() {
+		if err := cmManager.CleanupCMTemplates(); err != nil {
+			logger.WithError(err).Error("failed to clean up temporary template CM")
+		}
+	}()
 	if err := cmManager.CreateCMTemplates(); err != nil {
 		logger.WithError(err).Error("couldn't create template configMap")
-		gracefulExitWithConfigMapCleanup(o.noFail, "", cmManager)
+		return gracefulExit(false, failedSetupOutput)
```
petr-muller:
```suggestion
		return gracefulExit(o.noFail, failedSetupOutput)
```

droslean:
The PR is ok, for the current design. Of course, I wish, `os` package would have honored the `defers` when `os.Exit()`.

/lgtm

PR #111, opened 2019-03-15
in --local, create a PR number seeding from current time

droslean:
When the users are running the `pj-rehearse` tool with `--local`, creates a PR number instead of `0`. 
This change will make sure that the temporary configmaps won't all be deleted when multiple users are running with `--local`


petr-muller (approved):
/lgtm

PR #112, opened 2019-03-18
no need to configure template.Name.

droslean:
That is causing errors like [this](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_release/3132/rehearse-3132-pull-ci-openshift-console-master-e2e-aws-console/11)

ci-operator is generating the name depending on the filename already, and the file is mounted with a different name. 
So, if we configure a `template.Name` already, `ci-operator` will use that one instead. see [here](https://github.com/openshift/ci-operator/blob/master/cmd/ci-operator/main.go#L353-L361)

### DIFF
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -9,7 +9,6 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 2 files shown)


petr-muller (approved):
/lgtm

bbguimaraes:
Won't that affect other uses of the name, e.g. https://github.com/openshift/ci-operator-prowgen/blob/5e2ce4d7e073fe05a492a8a9d4850900d8e60f36/pkg/config/template.go#L96?

droslean:
I forgot to remove that as well.  Let me open a PR for this.

PR #114, opened 2019-03-18
remove unnecessary logging.

droslean:
That is a leftover when we first initialize the template differences.


petr-muller (approved):
/lgtm

PR #115, opened 2019-03-20
Add a script to validate Prowgen changes do not break openshift/release

petr-muller:
I was using a similar script for a while so I thought why not polish it a bit and make a presubmit out of it. This would remind us to always prepare followup PRs to `openshift/release` repo when we change Prowgen.

### DIFF
--- a/test/validate-prowgen-breaking-changes.sh
+++ b/test/validate-prowgen-breaking-changes.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+workdir="$( mktemp -d )"
+trap 'rm -rf "${workdir}"' EXIT
+
+git clone https://github.com/openshift/release.git --depth 1 "${workdir}/release"
+
+ci-operator-prowgen --from-dir "${workdir}/release/ci-operator/config" --to-dir "${workdir}/release/ci-operator/jobs"
+
+pushd "${workdir}/release"
+
+if [ -n "$(git status --porcelain)" ]; then
+  echo "[ERROR] Changes in openshift/release:"
+  git diff
+  echo "[ERROR] Running Prowgen in openshift/release results in changes ^^^"
+  echo "[ERROR] To avoid breaking openshift/release for everyone you should regenerate"
+  echo "[ERROR] the jobs there and merge the changes ASAP after this change to Prowgen"
+else
+  echo "Running Prowgen in openshift/release does not result in changes, no followups needed"
+fi
+
+popd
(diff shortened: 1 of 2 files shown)


test/validate-prowgen-breaking-changes.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+workdir="$( mktemp -d )"
+trap 'rm -rf "${workdir}"' EXIT
+
+git clone https://github.com/openshift/release.git --depth 1 "${workdir}/release"
```
stevekuznetsov, line 10: Can we make this optional and turn it off in CI since `clonerefs` does this for us?

test/validate-prowgen-breaking-changes.sh
```diff
@@ -0,0 +1,26 @@
...
+
+ci-operator-prowgen --from-dir "${workdir}/release/ci-operator/config" --to-dir "${workdir}/release/ci-operator/jobs"
+
+pushd "${workdir}/release"
+
+if [ -n "$(git status --porcelain)" ]; then
```
stevekuznetsov, line 16: Could also make this a `diff` a la the prowgen check script in release so the test image doesn't need `git`

stevekuznetsov (commented):
Sweet!

test/validate-prowgen-breaking-changes.sh
petr-muller: Hmm this should be run in presubmit for `openshift/ci-operator-prowgen` but needs to check out `openshift/release`. Is this possible to set up a presubmit like that? Somehow with `extra_refs`?

test/validate-prowgen-breaking-changes.sh
petr-muller: Could, but then I would need to make a pre-prowgen copy. I think we have git in the test image - if so, I'd like to keep this simple. If not, I'll change it.

test/validate-prowgen-breaking-changes.sh
stevekuznetsov: Yeah, totally - just add `openshift/release@master` to `extra_refs`

test/validate-prowgen-breaking-changes.sh
stevekuznetsov: I meant output to a different dir and diff the release master state but if we already have `git` then it's fine

test/validate-prowgen-breaking-changes.sh
bbguimaraes: That won't work for `ci-operator` tests.

stevekuznetsov: :man_facepalming:

bbguimaraes: It was also what I thought when I saw that line ;)  Too bad it doesn't work.

PR #116, opened 2019-03-20
prowgen: prune stale generated jobs

petr-muller:
Label the generated jobs so that Prowgen can recognize jobs it generated
in the past. When merging newly generated jobs into the old job
configuration, prune jobs that were generated earlier but were not
generated by the current run. This will prevent stale jobs being kept
in the release repo after the reason to generate the jobs is gone (for
example, when someone removes a clause in the `tests` array in their
ci-operator config, but forgets to remove the job that was generated for
it).

/cc @droslean @bbguimaraes @stevekuznetsov


cmd/ci-operator-prowgen/main.go
```diff
@@ -287,6 +287,7 @@ func generatePostsubmitForTest(
 	for k, v := range labels {
 		copiedLabels[k] = v
 	}
+	copiedLabels[jc.ProwJobLabelGenerated] = jc.Generated
```
stevekuznetsov, line 288: nit: the godoc is stale for this method :face_with_head_bandage:

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -285,6 +288,7 @@ func TestGeneratePodSpecTemplate(t *testing.T) {
 
 func TestGeneratePresubmitForTest(t *testing.T) {
 	newTrue := true
+	standardJobLabels := map[string]string{jobconfig.ProwJobLabelGenerated: jobconfig.Generated}
```
stevekuznetsov: Can we not use the imports here just so we test the actual values and someone can't break us by helpful refactors?

pkg/jobconfig/files.go
```diff
@@ -242,16 +248,60 @@ func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig) error
...
+			}
+		}
+		jobConfig.Postsubmits[repo] = jobs[:i]
+	}
+}
+
 // Given a JobConfig and a file path, write YAML representation of the config
 // to the file path. If the file already contains some jobs, new ones will be
-// merged with the existing ones.
+// merged with the existing ones. If the file contains
```
stevekuznetsov: TODO? ;)

stevekuznetsov (commented):
/lgtm
/hold
minor comments -- can we get a branch of prowgen before the promotion changes and use this feature to label jobs ?

petr-muller:
`ci/prow/breaking-changes — Job succeeded. `

Huh WTF, this should definitely be detected as a breaking change.

**Edit:** https://github.com/openshift/ci-operator-prowgen/pull/119 (oops)

pkg/jobconfig/files.go
```diff
@@ -242,16 +248,60 @@ func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig) error
...
+	for _, jobs := range jobConfig.Postsubmits {
+		for _, job := range jobs {
+			if _, isGenerated := job.Labels[ProwJobLabelGenerated]; isGenerated {
+				job.Labels[ProwJobLabelGenerated] = label
+			}
+		}
+	}
+}
+
+func pruneStaleGeneratedJobs(jobConfig *prowconfig.JobConfig, staleLabel string) {
```
droslean, line 268: I am not saying that this is wrong, but I personally try to avoid re-arranging the slice with the indexing. I don't trust it :)
How about:
```suggestion
func pruneStaleGeneratedJobs(jobConfig *prowconfig.JobConfig, staleLabel string) {
	for repo, jobs := range jobConfig.Presubmits {
		tempJobs := []prowconfig.Presubmit{}
		for _, job := range jobs {
			if label, isGenerated := job.Labels[ProwJobLabelGenerated]; !isGenerated || label != staleLabel {
				tempJobs = append(tempJobs, job)
			}
		}
		jobConfig.Presubmits[repo] = tempJobs
	}
	
	for repo, jobs := range jobConfig.Postsubmits {
		tempJobs := []prowconfig.Postsubmit{}
		for _, job := range jobs {
			if label, isGenerated := job.Labels[ProwJobLabelGenerated]; !isGenerated || label != staleLabel {
				tempJobs = append(tempJobs, job)
			}
		}
		jobConfig.Postsubmits[repo] = tempJobs
	}
}
```

droslean (commented):
Looks good in general.  Do we also need to update the integration test to check for `stale` labels?

petr-muller:
> minor comments -- can we get a branch of prowgen before the promotion changes and use this feature to label jobs ?

@stevekuznetsov uh, can you explain? I don't think I understand what you propose here.

cmd/ci-operator-prowgen/main.go
petr-muller: Will fix

cmd/ci-operator-prowgen/main_test.go
petr-muller: Sure.

pkg/jobconfig/files.go
petr-muller: Heh.

pkg/jobconfig/files.go
petr-muller: Yeah, what you suggest is what I would do normally, but I'm doing some Go reading and this is supposed to be more efficient (less allocations) and quite idiomatic Go way to filter out elements out of the slice...

@stevekuznetsov WDYT?

petr-muller:
> Do we also need to update the integration test to check for stale labels?

@droslean What do you propose to update? I added a [stale job to integration test data](https://github.com/openshift/ci-operator-prowgen/pull/116/files#diff-199824bc65444b5645f72cbf809a6abfR53), so our integration script already validates such job is removed

cmd/ci-operator-prowgen/main.go
petr-muller: Removed the godoc from both methods, did not say anything helpful beyond what the methods names suggest.

petr-muller:
Addressed all straightforward comments. I'm not sure what to do for https://github.com/openshift/ci-operator-prowgen/pull/116#issuecomment-475244149 and not yet decided if I want to do what Nikos suggests in https://github.com/openshift/ci-operator-prowgen/pull/116#discussion_r267771291

pkg/jobconfig/files.go
stevekuznetsov: Yeah I think the re-slicing (as originally implemented) is fine, pretty common in Go

stevekuznetsov:
I meant -- if we get this feature to cherry-pick on top of an older version of `prowgen` before we made promotions explicit and not implicit, we could then use a new prowgen version with this feature to remove the promotion jobs that only existed because we were generating them from implicit stanzas, right?

pkg/jobconfig/files.go
petr-muller: I'll keep it then.

petr-muller:
> I meant -- if we get this feature to cherry-pick on top of an older version of `prowgen` before we made promotions explicit and not implicit, we could then use a new prowgen version with this feature to remove the promotion jobs that only existed because we were generating them from implicit stanzas, right?

Ah! I think I found all of these (semi-manually) in https://github.com/openshift/release/pull/3182 but it's worth a shot, if the cherry-pick applies at least a little sanely so that we would not need to backport it.

stevekuznetsov:
Yeah we can try after this is in

/lgtm
/approve
/hold cancel
/woof
/meow

PR #117, opened 2019-03-20
Improve job selection on changed ci-op configs

droslean:
If only the `.tests` field in ci-op config changes, rehearse the corresponding jobs.

### DIFF
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -191,9 +191,9 @@ func GetPresubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfigs conf
 				}
 				if config.IsCiopConfigCM(env.ValueFrom.ConfigMapKeyRef.Name) {
 					if _, ok := ciopConfigs[env.ValueFrom.ConfigMapKeyRef.Key]; ok {
-
-						s := strings.Split(job.Name, "-")
-						testName := s[len(s)-1]
+						testName := strings.TrimPrefix(job.Name, "pull-ci-")
+						orgRepo := strings.Replace(repo, "/", "-", -1)
+						testName = strings.TrimPrefix(testName, fmt.Sprintf("%s-%s-", orgRepo, job.Brancher.Branches[0]))
 
 						affectedJob, ok := affectedJobs[env.ValueFrom.ConfigMapKeyRef.Key]
 						if ok && !affectedJob.Has(testName) {
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -8,9 +8,12 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	utildiff "k8s.io/apimachinery/pkg/util/diff"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
+
+	cioperatorapi "github.com/openshift/ci-operator/pkg/api"
 
 	"github.com/openshift/ci-operator-prowgen/pkg/config"
 )
@@ -38,11 +41,13 @@ const (
 	changedCiopConfigMsg = "ci-operator config file changed"
 )
 
-func GetChangedCiopConfigs(masterConfig, prConfig config.CompoundCiopConfig, logger *logrus.Entry) config.CompoundCiopConfig {
+func GetChangedCiopConfigs(masterConfig, prConfig config.CompoundCiopConfig, logger *logrus.Entry) (config.CompoundCiopConfig, map[string]sets.String) {
 	ret := config.CompoundCiopConfig{}
+	affectedJobs := map[string]sets.String{}
 
 	for filename, newConfig := range prConfig {
 		oldConfig, ok := masterConfig[filename]
+		jobs := sets.NewString()
 
 		// new ciop config
 		if !ok {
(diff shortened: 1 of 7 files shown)


pkg/diffs/diffs_test.go
```diff
@@ -81,7 +81,7 @@ func TestGetChangedCiopConfigs(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			before, after := tc.configGenerator()
-			actual := GetChangedCiopConfigs(before, after, logrus.NewEntry(logrus.New()))
+			actual, _ := GetChangedCiopConfigs(before, after, logrus.NewEntry(logrus.New()))
```
stevekuznetsov: If all the changes in `_test.go` files are to add `_` and `nil` I think we both know what's missing ;)

pkg/diffs/diffs.go
```diff
@@ -51,13 +57,36 @@ func GetChangedCiopConfigs(masterConfig, prConfig config.CompoundCiopConfig, log
 			continue
 		}
 
-		if !equality.Semantic.DeepEqual(oldConfig, newConfig) {
+		withoutTestsOldConfig := cioperatorapi.ReleaseBuildConfiguration{}
```
stevekuznetsov: There's a lot going on here -- can't we loop over the tests, and if none are found to be different, then use a top-level check?

```go
		if !equality.Semantic.DeepEqual(oldConfig, newConfig) {
			ret[filename] = newConfig
		}
		affectedJobs[filename] = sets.NewString()
		oldTests := getTestsByName(oldConfig.Tests)
		for as, test := range getTestsByName(newConfig.Tests) {
			if !equality.Semantic.DeepEqual(oldTests[as], test) {
				logger.WithField(logCiopConfig, filename).Info(changedCiopConfigMsg)
				affectedJobs[filename].Insert(as)
			}
		}
```

pkg/diffs/diffs_test.go
droslean: I [WIP] because of that. I will take care of it.

pkg/diffs/diffs.go
droslean: There is always a chance that the tests are changed but other fields as well. In that case, we want to rehearse all jobs.

stevekuznetsov: True!

stevekuznetsov (commented):
Is it possible to get this into the integration test, too?

droslean:
> Is it possible to get this into the integration test, too?

I will add them

pkg/diffs/diffs.go
```diff
@@ -51,13 +57,36 @@ func GetChangedCiopConfigs(masterConfig, prConfig config.CompoundCiopConfig, log
...
+		deepcopy.Copy(&withoutTestsOldConfig, masterConfig[filename])
+		withoutTestsOldConfig.Tests = nil
+
+		withoutTestsNewConfig := cioperatorapi.ReleaseBuildConfiguration{}
+		deepcopy.Copy(&withoutTestsNewConfig, prConfig[filename])
+		withoutTestsNewConfig.Tests = nil
```
petr-muller, line 62: We could avoid copying (and the external dep) if we saved tests, did the compare and put them back afterwards... Just a proposal for consideration, not a must-have.

```
oldTests := masterConfig[filename].Tests
masterConfig[filename].Tests = prConfig[filename].Tests // so that they are same
...compare...
masterConfig[filename].Tests = oldTests
```

pkg/diffs/diffs.go
```diff
@@ -166,6 +195,15 @@ func GetPresubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfigs conf
 				}
 				if config.IsCiopConfigCM(env.ValueFrom.ConfigMapKeyRef.Name) {
 					if _, ok := ciopConfigs[env.ValueFrom.ConfigMapKeyRef.Key]; ok {
+
+						s := strings.Split(job.Name, "-")
+						testName := s[len(s)-1]
```
petr-muller: I'm afraid this won't be enough, the test names may contain a `-` (`e2e-aws` and many more...), so the last `-`-separated fragment won't always be the test name...

petr-muller (commented):
I like the simplicity, I expected this to be more complex! I have one suggestion and found a possible bug - see the inline comments.

I'm also thinking we should add these detected affected jobs to the metrics structure, but I may do that in a followup.

pkg/diffs/diffs.go
droslean: Nice catch! I will address this. Thanks!

pkg/diffs/diffs.go
droslean: My intension here is actually to compare all the rest without the tests, since the tests are already saved in the original struct. My comparison holds a logic that if we found changes only in the `tests` field we should get the corresponding job is affected. 

If there are tests changed and some of the rest fields are changed also, we want to rehearse everything instead.

petr-muller: Yeah, I know - I just suggested how how do the same without copying the whole configs around.

droslean: We can always compare each field separately. I will check if this will be more efficient.

pkg/diffs/diffs.go
droslean: addressed

pkg/diffs/diffs.go
droslean: I could use the address of the value here.  addressed

petr-muller: This looks even better, cool!

pkg/diffs/diffs.go
```diff
@@ -166,6 +191,15 @@ func GetPresubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfigs conf
 				}
 				if config.IsCiopConfigCM(env.ValueFrom.ConfigMapKeyRef.Name) {
 					if _, ok := ciopConfigs[env.ValueFrom.ConfigMapKeyRef.Key]; ok {
+						testName := strings.TrimPrefix(job.Name, "pull-ci-")
+						orgRepo := strings.Replace(repo, "/", "-", -1)
+						testName = strings.TrimPrefix(testName, fmt.Sprintf("%s-%s-", orgRepo, job.Brancher.Branches[0]))
```
petr-muller, line 196: I am a little worried because `job.Brancher.Branches[0]` is a regex, but we generate presubmits as plain names anyway...

petr-muller (approved):
/lgtm

PR #118, opened 2019-03-20
rehearse: detect cluster profile changes

bbguimaraes:
- [x] Changes to cluster profiles are identified.
- [x] Jobs that use those profiles are chosen for rehearsal.
- [x] Integration test verifies the above (see note below).
- [x] `ConfigMap` manager is changed to also create cluster profiles.
- [x] Jobs are changed to use those `CM`s.
- [x] More thorough unit testing.

Follow-up:
- Integration tests with `--allow-volumes` (potentially fixed in https://github.com/openshift/ci-operator-prowgen/pull/122).
- ConfigMap create/update method ([comment](https://github.com/openshift/ci-operator-prowgen/pull/118/#discussion_r268263411), https://github.com/openshift/ci-operator-prowgen/pull/123).
- Load plugin config from Prow and use the updateconfig library to generate the CMs ([comment](https://github.com/openshift/ci-operator-prowgen/pull/118/#discussion_r268264133)).

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -4,6 +4,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -193,6 +194,15 @@ func rehearseMain() int {
 		changedTemplates = diffs.GetChangedTemplates(masterConfig.Templates, prConfig.Templates, logger)
 		metrics.RecordChangedTemplates(changedTemplates)
 	}
+	changedClusterProfiles, err := config.GetChangedClusterProfiles(o.releaseRepoPath, jobSpec.Refs.BaseSHA)
+	if err != nil {
+		logger.WithError(err).Error("could not get cluster profile differences")
+		return gracefulExit(o.noFail, misconfigurationOutput)
+	}
+	if len(changedClusterProfiles) != 0 {
+		logger.WithField("profiles", changedClusterProfiles).Info("cluster profiles changed")
+		metrics.RecordChangedClusterProfiles(changedClusterProfiles)
+	}
 
 	namespace := prConfig.Prow.ProwJobNamespace
 	if o.local {
@@ -213,6 +223,10 @@ func rehearseMain() int {
 	}()
 	if err := cmManager.CreateCMTemplates(); err != nil {
 		logger.WithError(err).Error("couldn't create template configMap")
+		return gracefulExit(o.noFail, failedSetupOutput)
+	}
+	if err := cmManager.CreateClusterProfiles(filepath.Join(o.releaseRepoPath, config.ClusterProfilesPath), changedClusterProfiles); err != nil {
+		logger.WithError(err).Error("couldn't create cluster profile ConfigMaps")
 		return gracefulExit(o.noFail, failedSetupOutput)
 	}
 
@@ -242,8 +256,11 @@ func rehearseMain() int {
 	presubmitsWitchChangedCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger, affectedJobs)
 	metrics.RecordOpportunity(presubmitsWitchChangedCiopConfigs, "ci-operator-config-change")
 	toRehearse.AddAll(presubmitsWitchChangedCiopConfigs)
+	toRehearseClusterProfiles := diffs.GetPresubmitsForClusterProfiles(prConfig.Prow, changedClusterProfiles, logger)
+	metrics.RecordOpportunity(toRehearseClusterProfiles, "cluster-profile-change")
+	toRehearse.AddAll(toRehearseClusterProfiles)
 
-	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates)
+	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
 	metrics.RecordActual(rehearsals)
 	if len(rehearsals) == 0 {
 		logger.Info("no jobs to rehearse have been found")
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -22,6 +22,8 @@ const (
 	CiopConfigInRepoPath = "ci-operator/config"
 	// TemplatesPath is the path of the templates from release repo
 	TemplatesPath = "ci-operator/templates"
+	// ClusterProfilesPath is where profiles are stored in the release repo
+	ClusterProfilesPath = "cluster/test-deploy"
 )
 
 // ReleaseRepoConfig contains all configuration present in release repo (usually openshift/release)
@@ -141,4 +143,18 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*R
 	}
 
 	return config, nil
+}
+
+// GetChangedClusterProfiles returns the name and a hash of the contents of all
+// cluster profiles that changed since the revision `baseRev`.
+func GetChangedClusterProfiles(path, baseRev string) (ret []ClusterProfile, err error) {
+	// Sample output (with abbreviated hashes) from git-diff-tree(1):
+	// :100644 100644 bcd1234 0123456 M file0
+	cmd := []string{"diff-tree", "--diff-filter=ABCMRTUX", baseRev + ":" + ClusterProfilesPath, "HEAD:" + ClusterProfilesPath}
+	if diff, err := git(path, cmd...); err == nil && diff != "" {
+		for _, l := range strings.Split(strings.TrimSpace(diff), "\n") {
+			ret = append(ret, ClusterProfile{Name: l[99:], TreeHash: l[56:96]})
+		}
+	}
+	return
 }
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -31,15 +31,22 @@ type ReleaseRepoConfig struct {
 	Templates  CiTemplates
 }
 
-func revParse(repoPath string, args ...string) (string, error) {
-	cmd := exec.Command("git", append([]string{"rev-parse"}, args...)...)
+func git(repoPath string, args ...string) (string, error) {
+	cmd := exec.Command("git", args...)
 	cmd.Dir = repoPath
-	sha, err := cmd.Output()
+	out, err := cmd.CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("'%s' failed with error=%v", cmd.Args, err)
+		return "", fmt.Errorf("'%s' failed with error=%v, output:\n%s", cmd.Args, err, out)
 	}
+	return string(out), nil
+}
 
-	return strings.TrimSpace(string(sha)), nil
+func revParse(repoPath string, args ...string) (string, error) {
+	out, err := git(repoPath, append([]string{"rev-parse"}, args...)...)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(out), nil
 }
 
 func gitCheckout(candidatePath, baseSHA string) error {
--- a/pkg/config/release_test.go
+++ b/pkg/config/release_test.go
@@ -1,0 +1,70 @@
+package config
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestGetChangedClusterProfiles(t *testing.T) {
+	dir, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+	profilesPath := filepath.Join(dir, ClusterProfilesPath)
+	for _, f := range []string{
+		"nochanges/file", "changeme/file", "removeme/file", "moveme/file",
+		"renameme/file", "dir/dir/file",
+	} {
+		path := filepath.Join(dir, ClusterProfilesPath, f)
+		if err := os.MkdirAll(filepath.Dir(path), 0775); err != nil {
+			t.Fatal(err)
+		}
+		if err := ioutil.WriteFile(path, []byte(f+"content"), 0664); err != nil {
+			t.Fatal(err)
+		}
+	}
+	p := exec.Command("sh", "-ec", fmt.Sprintf(`
+git init --quiet .
+git config user.name test
+git config user.email test
+git add .
+git commit --quiet -m initial
+cd %s
+> changeme/file
+git rm --quiet removeme/file
+mkdir new/ renamed/
+> new/file
+git add new/file
+git mv moveme/file moveme/moved
+git mv renameme/file renamed/file
+> dir/dir/file
+git commit --quiet -am changes
+git rev-parse HEAD^
+`, profilesPath))
+	p.Dir = dir
+	out, err := p.CombinedOutput()
+	if err != nil {
+		t.Fatalf("%q failed, output:\n%s", p.Args, out)
+	}
+	changed, err := GetChangedClusterProfiles(dir, strings.TrimSpace(string(out)))
+	if err != nil {
+		t.Fatal(err)
+	}
+	expected := []ClusterProfile{
+		{Name: "changeme", TreeHash: "df2b8fc99e1c1d4dbc0a854d9f72157f1d6ea078"},
+		{Name: "dir", TreeHash: "b4c3cc91598b6469bf7036502b8ca2bd563b0d0a"},
+		{Name: "moveme", TreeHash: "03b9d461447abb84264053a440b4c715842566bb"},
+		{Name: "new", TreeHash: "df2b8fc99e1c1d4dbc0a854d9f72157f1d6ea078"},
+		{Name: "renamed", TreeHash: "9bbab5dcf83793f9edc258136426678cccce940e"},
+	}
+	if !reflect.DeepEqual(changed, expected) {
+		t.Fatalf("want %s, got %s", expected, changed)
+	}
+}
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -85,6 +85,18 @@ func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logg
 	}
 }
 
+func (c *TemplateCMManager) createCM(cm *v1.ConfigMap) error {
+	if cm.ObjectMeta.Labels == nil {
+		cm.ObjectMeta.Labels = map[string]string{}
+	}
+	cm.ObjectMeta.Labels[createByRehearse] = "true"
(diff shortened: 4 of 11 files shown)


bbguimaraes:
> `pull-ci-openshift-ci-operator-prowgen-master-unit`
> `--- FAIL: TestIntegration (0.07s)`

/me confused

bbguimaraes:
Only a few minor items left. The existing bits are unlikely to change, so this is in a reviewable state. I've removed the integration test bits for now to reduce the noise.

pkg/config/release.go
```diff
@@ -135,3 +143,16 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*R
...
 }
+
+func GetChangedClusterProfiles(path, baseSHA string) (ret []ClusterProfile, err error) {
+	diff, err := git(path, "diff-tree", "--diff-filter=d", baseSHA+":"+ClusterProfilesPath, "HEAD:"+ClusterProfilesPath)
+	if err != nil || diff == "\n" {
+		return
+	}
+	for _, l := range strings.Split(strings.TrimSpace(diff), "\n") {
+		var _s, hash, name string
+		fmt.Sscanf(l, ":%6s %6s %40s %40s %s\t%s\t", &_s, &_s, &_s, &hash, &_s, &name)
```
droslean: oh jesus. At least can you try to add a comment for this line?

pkg/config/template.go
```diff
@@ -85,6 +88,22 @@ func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logg
 	}
 }
 
+func (c *TemplateCMManager) createCM(cm *v1.ConfigMap) error {
+	l := &cm.ObjectMeta.Labels
```
droslean: I think you don't need `&` here. Also, you will be able to update the value below without passing the pointer.  However, you can also `make()` it.

pkg/diffs/diffs.go
```diff
@@ -175,3 +175,35 @@ func GetPresubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfigs conf
...
+				}
+				for _, s := range v.Projected.Sources {
+					if s.ConfigMap == nil {
+						continue
+					}
+					n := strings.TrimPrefix(s.ConfigMap.Name, "cluster-profile-")
+					for _, p := range profiles {
+						if n == p.Name {
+							ret.Add(repo, job)
+							break top
```
droslean: This is not necessary

droslean (commented):
Nice start, But I don't like the idea of getting the cluster profile differences from `git diff`. I would prefer just comparing the bytes of the file. Also, for the cm name hash, we can get the hash from the cluster-profile's data.

pkg/config/release.go
bbguimaraes: We execute `git diff-tree` then parse its well-documented output to extract the hash and name from each line. What comment would you add that wouldn't be parroting what the code already says?

pkg/config/template.go
bbguimaraes: Would the assignment inside the `if` also work if it weren't a pointer?

pkg/diffs/diffs.go
bbguimaraes: Not strictly necessary, no, but why bother going through the rest of the iterations if we already know the answer?

pkg/config/template.go
droslean: it should

pkg/config/release.go
bbguimaraes: It *would* be nice to ignore the values we don't care about (à la c's `*` modifier) instead of using the dummy `_s` variable, but I don't think go supports that.

pkg/diffs/diffs.go
droslean: you will skip only `			if job.Agent != string(pjapi.KubernetesAgent) {
				continue
			}` right?

pkg/config/template.go
bbguimaraes: Doesn't look like it (which follows my understanding of how this should work, but there might be some `map` magic I am not aware of):

```go
package main

import "fmt"

func main() {
    type S struct {m map[int]int}
    s := S{}
    n := s.m
    fmt.Println(n == nil)
    if n == nil {
        n = map[int]int{}
    }
    n[0] = 1
    fmt.Printf("%#v %#v\n", s.m, n)
    t := S{}
    p := &t.m
    fmt.Println(*p == nil)
    if *p == nil {
        *p = map[int]int{}
    }
    (*p)[0] = 1
    fmt.Printf("%#v %#v\n", t.m, p)
}
```

```
true
map[int]int(nil) map[int]int{0:1}
true
map[int]int{0:1} &map[int]int{0:1}
```

pkg/diffs/diffs.go
bbguimaraes: No, `break top` will break from all `for` loops as soon as we find that the job matches our search criteria. Removing the break would cause it to unnecessarily look at all the remaing volumes, sources, and profiles. It would also add duplicates in the (unlikely) case where a job uses the profile more than once or more than one profile.

bbguimaraes: All `for` loops under the `top` label, that is.

bbguimaraes:
Fixed an issue with the version of `git` we use in CI (`1.8.3.1`, which is now *six years* old) not recognizing lower-case letters as exclusion filters in `--diff-filter`.

cmd/pj-rehearse/main.go
```diff
@@ -191,6 +192,15 @@ func rehearseMain() int {
...
 		metrics.RecordChangedTemplates(changedTemplates)
 	}
+	changedClusterProfiles, err := config.GetChangedClusterProfiles(o.releaseRepoPath, jobSpec.Refs.BaseSHA)
+	if err != nil {
+		logger.WithError(err).Error("could not get cluster profile differences")
+		return gracefulExit(o.noFail, misconfigurationOutput)
+	}
+	if len(changedClusterProfiles) != 0 {
+		logger.WithField("profiles", changedClusterProfiles).Info("cluster profiles changed")
+		metrics.RecordChangedClusterProfiles(changedClusterProfiles)
```
petr-muller, line 204: :heart:

pkg/rehearse/metrics.go
```diff
@@ -21,9 +21,10 @@ type Metrics struct {
 	Repo string `json:"repo"`
 	Pr   int    `json:"pr"`
 
-	ChangedCiopConfigs []string `json:"changed_ciop_configs"`
-	ChangedPresubmits  []string `json:"changed_presubmits"`
-	ChangedTemplates   []string `json:"changed_templates"`
+	ChangedCiopConfigs     []string                `json:"changed_ciop_configs"`
+	ChangedPresubmits      []string                `json:"changed_presubmits"`
+	ChangedTemplates       []string                `json:"changed_templates"`
+	ChangedClusterProfiles []config.ClusterProfile `json:"changed_cluster_profiles"`
```
petr-muller: Can we store just names here to keep the output easier to eyball?

bbguimaraes: Ah, `ClusterProfile` used to be just `type ClusterProfile string`, I forgot to change it when I added the other field.

pkg/config/template.go
```diff
@@ -85,6 +88,22 @@ func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logg
...
+	if *l == nil {
+		*l = map[string]string{}
+	}
+	(*l)[createByRehearse] = "true"
```
stevekuznetsov: Why can't we:

```go
if cm.ObjectMeta.Labels == nil {
    cm.ObjectMeta.Labels = map[string]string{}
}
cm.ObjectMeta.Labels[createByRehearse] = "true"
cm.ObjectMeta.Labels[rehearseLabelPull] = strconv.Itoa(c.prNumber)
```

pkg/config/template.go
```diff
@@ -85,6 +88,22 @@ func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logg
...
+	(*l)[rehearseLabelPull] = strconv.Itoa(c.prNumber)
+	if _, err := c.cmclient.Create(cm); err == nil || !kerrors.IsAlreadyExists(err) {
+		return err
+	}
+	if _, err := c.cmclient.Update(cm); err != nil {
```
stevekuznetsov: We want to `Get` before we `Update` and make sure to do retries to ensure we don't get into the odd version issues

pkg/config/template.go
```diff
@@ -97,28 +116,58 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
...
+	ret.ObjectMeta = metav1.ObjectMeta{
+		Name: fmt.Sprintf("rehearse-cluster-profile-%s-%s", profile.Name, profile.TreeHash[:5]),
+	}
+	ret.Data = map[string]string{}
+	profilePath := filepath.Join(dir, profile.Name)
+	err := filepath.Walk(profilePath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if !info.IsDir() {
```
stevekuznetsov: nit: Go idiom is:

```go
if info.IsDir() {
    return nil
}
```

to reduce code at high levels of indentation

pkg/config/template.go
```diff
@@ -97,28 +116,58 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
...
+			if b, err := ioutil.ReadFile(path); err == nil {
+				ret.Data[strings.TrimPrefix(path, profilePath)[1:]] = string(b)
```
stevekuznetsov: Why `[1:]`?

pkg/config/template.go
```diff
@@ -97,28 +116,58 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
...
+		}
+		c.logger.WithFields(logrus.Fields{"cluster-profile": cm.ObjectMeta.Name}).Info("creating rehearsal cluster profile ConfigMap")
+		if err := c.createCM(cm); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return kutilerrors.NewAggregate(errs)
+}
+
+func genClusterProfileCM(dir string, profile ClusterProfile) (*v1.ConfigMap, error) {
```
stevekuznetsov, line 143: Could we load plugin config from Prow and use the updateconfig library to do this?

pkg/diffs/diffs.go
stevekuznetsov: Named breaks are a code smell -- try to break out the inner loops into functions and keep the indentation low, helps to test as well

pkg/config/release.go
```diff
@@ -135,3 +143,16 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*R
 
 	return config, nil
 }
+
+func GetChangedClusterProfiles(path, baseSHA string) (ret []ClusterProfile, err error) {
```
stevekuznetsov: For the other values we load and diff the deserialized values -- pros/cons for doing this differently?

pkg/config/template.go
bbguimaraes: I wrote the exact same code as a comparison when discussing this with @droslean.

Personal preference, I guess, but it's 2x1 now. Changed.

pkg/config/template.go
bbguimaraes: I merely extracted this to a separate function without changing the behavior (i.e. this is what we currently do in production), but OK.

pkg/config/template.go
bbguimaraes: Changed.

pkg/config/template.go
bbguimaraes: `/`. Changed to `filepath.Base`.

pkg/config/template.go
bbguimaraes: Likely, I'll check that.

pkg/diffs/diffs.go
bbguimaraes: Generally, I agree, but this is the dumbest of loops. Extracted the relevant part to a local lambda function.

pkg/config/release.go
bbguimaraes: Pros:

- We don't need the actual contents of the files until we create the ConfigMaps.
- Even then, we only need the contents of the profiles that changed (none or one, 99% of the time).
- No need to `checkout` one version, then `checkout` back again (if we weren't already doing it, of course).
- No need to compare all files manually.
- No need to invent some ad-hoc hashing algorithm, which has to handle multiple files.

Cons:

- There was one minor issue with the `git` version we use in CI, that was caught by the unit test.
- Uses a semi-obscure, low-level `git` command.
- ?

Essentially, the problem is: which files changed from one revision to the other, and I think `git` is the appropriate tool for that. The fact that it already stores a hash of the contents of the directories is a very nice bonus.  We replace a lot of complicated steps with an algorithm that is essentially `exec+scanf`.

Plus, `git` is awesome =)

pkg/config/release.go
stevekuznetsov: Document your `_s` idea, it's not obvious

pkg/config/release.go
stevekuznetsov: Another con is we end up finding changes in different ways in one code-base. @petr-muller you have any objections to that?

pkg/config/template.go
stevekuznetsov: We can handle it in a follow-up, then :)

droslean: I think we can remove the `Update`. Since we are hasing the name of the cm.

pkg/config/template.go
```diff
@@ -97,30 +115,58 @@ func (c *TemplateCMManager) CreateCMTemplates() error {
...
+	ret := &v1.ConfigMap{}
+	ret.ObjectMeta = metav1.ObjectMeta{
+		Name: fmt.Sprintf("rehearse-cluster-profile-%s-%s", profile.Name, profile.TreeHash[:5]),
+	}
+	ret.Data = map[string]string{}
```
droslean, line 148: is this really needed?

pkg/diffs/diffs.go
```diff
@@ -217,3 +217,39 @@ func getTestsByName(tests []cioperatorapi.TestStepConfiguration) map[string]ciop
...
+			if v.Name != "cluster-profile" || v.Projected == nil {
+				continue
+			}
+			for _, s := range v.Projected.Sources {
+				if s.ConfigMap == nil {
+					continue
+				}
+				n := strings.TrimPrefix(s.ConfigMap.Name, "cluster-profile-")
+				for _, p := range profiles {
+					if n == p.Name {
```
droslean, line 238: since we need only the names, why not passing a `sets.String` instead of `[]config.ClusterProfile`

pkg/config/template.go
bbguimaraes: Isn't it `nil` when `ret` is initialized with `v1.ConfigMap{}`?

pkg/diffs/diffs.go
bbguimaraes: It would complicate the caller, give us only a very minimal performance gain (1/15th of the total execution time of the function, in my most pessimistic and unrealistic profile) and not make the code much clearer, IMO.

pkg/config/release.go
bbguimaraes: So, it turns out the strings we are interested in appear at fixed offsets, so we can just index into the input string directly. I tried to think of every possible transformation and added them to the unit tests, the algorithm worked for all of them.

pkg/rehearse/metrics.go
bbguimaraes: Forgot to mention: changed in 323f9d02f0612bf84e219de2e65edefafa4d6619.

pkg/config/release.go
bbguimaraes: Cluster profiles have some unique characteristics that make this approach more beneficial. Not all of those items are true for the other cases, e.g. the first is not true for `ci-operator` configuration files, the last, for templates (although we could probably use this as a first pass in those cases as well).

The latest version of the code is nine lines, 431 characters long. Looking at the code we have for the other cases (and keeping in mind cluster profiles would require special treatment), it seems to be worth having a separate loading mechanism.

petr-muller: I was thinking about this when I reviewed the PR, and at first I was thinking, "why do we need a second path doing the same", but ended thinking what Bruno did is really better, and the duplicate way is a reasonable price. I think we could migrate detecting changes in tempaltes to this way too, but I think that could be a tech debt-paying followup.

pkg/config/release.go
```diff
@@ -21,7 +21,8 @@ const (
 	// CiopConfigInRepoPath is the ci-operator config path from release repo
 	CiopConfigInRepoPath = "ci-operator/config"
 	// TemplatesPath is the path of the templates from release repo
-	TemplatesPath = "ci-operator/templates"
+	TemplatesPath       = "ci-operator/templates"
+	ClusterProfilesPath = "cluster/test-deploy"
```
stevekuznetsov, line 26: GoDoc please ;)

pkg/config/release.go
```diff
@@ -135,3 +143,13 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*R
 
 	return config, nil
 }
+
+func GetChangedClusterProfiles(path, baseSHA string) (ret []ClusterProfile, err error) {
```
stevekuznetsov: Could we get GoDoc here with an example line of output from `git`?

pkg/config/release_test.go
```diff
@@ -0,0 +1,70 @@
...
+	} {
+		path := filepath.Join(dir, ClusterProfilesPath, f)
+		if err := os.MkdirAll(filepath.Dir(path), 0775); err != nil {
+			t.Fatal(err)
+		}
+		if err := ioutil.WriteFile(path, []byte(f+"content"), 0664); err != nil {
+			t.Fatal(err)
+		}
+	}
+	p := exec.Command("sh", "-ec", fmt.Sprintf(`
```
stevekuznetsov, line 33: Pretty crazy but nice

pkg/config/template.go
stevekuznetsov: Also in a follow-up so this can be just a move

stevekuznetsov (commented):
Everything here looks pretty good -- what's left?

bbguimaraes:
Updated.

> @stevekuznetsov
> Everything here looks pretty good -- what's left?

See the description. I separated the three conditions that already existed before this PR. Those do not interfere with the code here and should be dealt with in separate PRs, IMO.

pkg/config/template.go
bbguimaraes: That looks promising. I think it should be an effort that covered the other files we load as well (templates for sure, ci-operator configuration files maybe), so I added it as a follow-up.

pkg/config/release_test.go
bbguimaraes: -- my epitaph

petr-muller (approved):
/lgtm
/hold

What's left for this to be WIP? Even your description checklist says it's complete and I agree the remaining items can be done as followups.

To me it looks good (btw I really like your code, reads nice and I like the fact that adding a feature are 99% additions and not changes).

bbguimaraes:
Why, thank you =)

The WIP was just to give people a chance to check the commits addressing the latest review. I will squash the history and remove the label.

stevekuznetsov:
/hold cancel
/lgtm
:shipit:

PR #122, opened 2019-03-25
On a template change, rehearse all cluster types.

droslean:
When a template file will change, rehearse all cluster types by choosing a random job that uses the corresponding template and cluster type.
Ignores the cluster types of jobs from the changed presubmits that uses the template.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -38,6 +38,8 @@ const (
 	logRehearsalJob              = "rehearsal-job"
 	logCiopConfigFile            = "ciop-config-file"
 	logCiopConfigRepo            = "ciop-config-repo"
+
+	clusterTypeEnvName = "CLUSTER_TYPE"
 )
 
 // Loggers holds the two loggers that will be used for normal and debug logging respectively.
@@ -203,17 +205,14 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 				continue
 			}
 
-			exists, index, templateKey := hasChangedTemplateVolume(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates)
-			if exists {
-				if templateData, err := config.GetTemplateData(templates[templateKey]); err != nil {
-					jobLogger.WithError(err).WithField("template-name", templates[templateKey].Name).Warn("couldn't get template's data. Job won't be rehearsed")
+			if allowVolumes {
+				if err := replaceCMTemplateName(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates); err != nil {
+					jobLogger.WithError(err).Warn("Failed to replace the configmap name to the temporary one")
 					continue
-				} else {
-					templateName := config.GetTemplateName(templateKey)
-					rehearsal.Spec.Volumes[index].VolumeSource.ConfigMap.Name = config.GetTempCMName(templateName, templateKey, templateData)
 				}
+
+				replaceClusterProfiles(rehearsal.Spec.Volumes, profiles, loggers.Debug.WithField("name", job.Name))
 			}
-			replaceClusterProfiles(rehearsal.Spec.Volumes, profiles, loggers.Debug.WithField("name", job.Name))
 
 			jobLogger.WithField(logRehearsalJob, rehearsal.Name).Info("Created a rehearsal job to be submitted")
 			rehearsals = append(rehearsals, rehearsal)
@@ -223,24 +222,93 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 	return rehearsals
 }
 
-func hasChangedTemplateVolume(volumeMounts []v1.VolumeMount, volumes []v1.Volume, templates config.CiTemplates) (bool, int, string) {
-	var templateKey string
-	var volumeName string
+// AddRandomJobsForChangedTemplates finds jobs from the PR config that are using a specific template with a specific cluster type.
+// The job selection is done by iterating in an unspecified order, which avoids picking the same job
+// So if a template will be changed, find the jobs that are using a template in combination with the `aws`,`openstack`,`gcs` and `libvirt` cluster types.
+func AddRandomJobsForChangedTemplates(templates config.CiTemplates, prConfigPresubmits map[string][]prowconfig.Presubmit, loggers Loggers, prNumber int) config.Presubmits {
+	rehearsals := make(config.Presubmits)
 
+	for templateFile := range templates {
+		for _, clusterType := range []string{"aws", "gcs", "openstack", "libvirt"} {
+			if repo, job := pickTemplateJob(prConfigPresubmits, templateFile, clusterType); job != nil {
+				jobLogger := loggers.Job.WithFields(logrus.Fields{"target-repo": repo, "target-job": job.Name})
+				jobLogger.Info("Picking job to rehearse the template changes")
+				rehearsals[repo] = append(rehearsals[repo], *job)
+			}
+		}
+	}
+	return rehearsals
+}
+
+func replaceCMTemplateName(volumeMounts []v1.VolumeMount, volumes []v1.Volume, templates config.CiTemplates) error {
+	replace := func(v *v1.Volume) error {
+		volumeName, templateKey := hasChangedTemplateVolume(volumeMounts, *v, templates)
+		if len(volumeName) == 0 || len(templateKey) == 0 || volumeName != v.Name {
+			return nil
+		}
+
+		templateData, err := config.GetTemplateData(templates[templateKey])
+		if err != nil {
+			return fmt.Errorf("couldn't get template's data: %s: %v", templates[templateKey], err)
+		}
+
+		name := config.GetTempCMName(config.GetTemplateName(templateKey), templateKey, templateData)
+		v.VolumeSource.ConfigMap.Name = name
+
+		return nil
+	}
+
+	for _, volume := range volumes {
+		if err := replace(&volume); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func hasChangedTemplateVolume(volumeMounts []v1.VolumeMount, volume v1.Volume, templates config.CiTemplates) (string, string) {
 	for _, volumeMount := range volumeMounts {
 		if _, ok := templates[volumeMount.SubPath]; ok {
-			templateKey = volumeMount.SubPath
-			volumeName = volumeMount.Name
+			return volumeMount.Name, volumeMount.SubPath
 		}
 	}
+	return "", ""
+}
 
-	for index, volume := range volumes {
-		if volume.Name == volumeName {
-			return true, index, templateKey
+func pickTemplateJob(presubmits map[string][]prowconfig.Presubmit, templateFile, clusterType string) (string, *prowconfig.Presubmit) {
+	for repo, jobs := range presubmits {
+		for _, job := range jobs {
+			if job.Agent != string(pjapi.KubernetesAgent) {
+				continue
+			}
+
+			if hasClusterType(job, clusterType) && hasTemplateFile(job, templateFile) {
+				return repo, &job
+			}
 		}
 	}
+	return "", nil
+}
 
-	return false, 0, ""
+func hasClusterType(job prowconfig.Presubmit, clusterType string) bool {
+	for _, env := range job.Spec.Containers[0].Env {
+		if env.Name == clusterTypeEnvName && env.Value == clusterType {
+			return true
+		}
+	}
+	return false
+}
+
+func hasTemplateFile(job prowconfig.Presubmit, templateFile string) bool {
+	if job.Spec.Containers[0].VolumeMounts != nil {
+		for _, volumeMount := range job.Spec.Containers[0].VolumeMounts {
+			if volumeMount.SubPath == templateFile {
+				return true
+			}
+		}
+	}
+	return false
 }
 
 func replaceClusterProfiles(volumes []v1.Volume, profiles []config.ClusterProfile, logger *logrus.Entry) {
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -253,9 +253,14 @@ func rehearseMain() int {
 	metrics.RecordChangedPresubmits(toRehearse)
 	metrics.RecordOpportunity(toRehearse, "direct-change")
 
-	presubmitsWitchChangedCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger, affectedJobs)
-	metrics.RecordOpportunity(presubmitsWitchChangedCiopConfigs, "ci-operator-config-change")
-	toRehearse.AddAll(presubmitsWitchChangedCiopConfigs)
+	presubmitsWithChangedCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigs, logger, affectedJobs)
+	metrics.RecordOpportunity(presubmitsWithChangedCiopConfigs, "ci-operator-config-change")
+	toRehearse.AddAll(presubmitsWithChangedCiopConfigs)
+
+	presubmitsWithChangedTemplates := rehearse.AddRandomJobsForChangedTemplates(changedTemplates, prConfig.Prow.JobConfig.Presubmits, loggers, prNumber)
+	metrics.RecordOpportunity(presubmitsWithChangedTemplates, "templates-change")
+	toRehearse.AddAll(presubmitsWithChangedTemplates)
+
 	toRehearseClusterProfiles := diffs.GetPresubmitsForClusterProfiles(prConfig.Prow, changedClusterProfiles, logger)
 	metrics.RecordOpportunity(toRehearseClusterProfiles, "cluster-profile-change")
 	toRehearse.AddAll(toRehearseClusterProfiles)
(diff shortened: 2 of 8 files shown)


pkg/rehearse/jobs.go
```diff
@@ -219,9 +230,82 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 		}
 	}
 
+	// for the remaining cluster types of each template, pick a `random` corresponding job to rehearse.
+	for templateFile, clusterTypes := range templatesWithClusterTypes {
+		for clusterType := range clusterTypes {
+			if job := pickTemplateJob(prConfig.JobConfig.Presubmits, templateFile, clusterType); job != nil {
+
+				// change the configmap name with the temporary created one.
```
stevekuznetsov: This seems mostly duped from above?

pkg/rehearse/jobs.go
```diff
@@ -219,9 +230,82 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+		}
+	}
+
 	return rehearsals
 }
 
+func getTemplatesWithClusterTypes(templates config.CiTemplates) map[string]sets.String {
+	templatesKeys := make(map[string]sets.String)
+	for filename := range templates {
+		templatesKeys[filename] = sets.NewString([]string{"aws", "gcs", "openstack", "libvirt"}...)
```
stevekuznetsov: We already have this in `prowgen` right? Need to merge repos ...

pkg/rehearse/jobs.go
```diff
@@ -219,9 +230,82 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+	for _, jobs := range presubmits {
+		for _, job := range jobs {
+			if job.Agent != string(pjapi.KubernetesAgent) {
+				continue
+			}
+
+			if hasClusterType(job, clusterType) {
+				if ok := hasTemplateFile(job, templateFile); ok {
+					for index, env := range job.Spec.Containers[0].Env {
+						if env.Name == testCommandEnvName {
```
stevekuznetsov: So what happens if we don't have a TEST_COMMAND env?

stevekuznetsov (commented):
I'm still not super certain this is a good idea -- what information does this approach give us that running the job without changing it's command does not? We get a strict subset of information, right? Why not just run the job fully?

pkg/rehearse/jobs.go
droslean: Yea shortly.  I will put the logic in a different function.

droslean:
> I'm still not super certain this is a good idea -- what information does this approach give us that running the job without changing it's command does not? We get a strict subset of information, right? Why not just run the job fully?

@stevekuznetsov Double thinking that, you are right. I guess is not a big deal to run the whole job. I will address that.

pkg/rehearse/jobs.go
droslean: we don't have them anywhere. But we could create them and use them both in prowgen and pj-rehearse.

pkg/rehearse/jobs.go
droslean: removed.

pkg/rehearse/jobs.go
```diff
@@ -203,25 +209,114 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 				continue
 			}
 
-			exists, index, templateKey := hasChangedTemplateVolume(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates)
+			exists, index, cmName, err := getVolumeIndexAndTempCMName(rehearsal.Spec.Containers[0].VolumeMounts, rehearsal.Spec.Volumes, templates)
```
petr-muller: :+1: for this refactor

pkg/rehearse/jobs.go
```diff
@@ -203,25 +209,114 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+				rehearsal.Spec.Volumes[index].VolumeSource.ConfigMap.Name = cmName
 			}
 
 			jobLogger.WithField(logRehearsalJob, rehearsal.Name).Info("Created a rehearsal job to be submitted")
 			rehearsals = append(rehearsals, rehearsal)
 		}
 	}
 
+	// for the remaining cluster types of each template, pick a `random` corresponding job to rehearse.
+	for templateFile, clusterTypes := range templatesWithClusterTypes {
```
petr-muller: I really think this should go into a separate method and not into `ConfigureRehearsalJobs`. Let's keep CRJ's concern to turning jobs into rehearsals, it's already quite big. Let the new method return `config.Presubmits` and merge them at the call site. The new method would be also more testable.

pkg/rehearse/jobs.go
```diff
@@ -203,25 +209,114 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+
+				rehearsals = append(rehearsals, rehearsal)
+			}
+		}
+	}
+
 	return rehearsals
 }
 
+func getVolumeIndexAndTempCMName(volumeMounts []v1.VolumeMount, volumes []v1.Volume, templates config.CiTemplates) (bool, int, string, error) {
```
petr-muller: Returning four items is pushing readability limits. Maybe we could do:

```
type VolumeRef struct {
  index int
  name string
}
```
and let the function return `(*VolumeRef, error)` where `nil` means not found.

Just an idea, not a must address.

droslean: I was thinking that too, but I didn't do it, because it's being used only 2 times for the templates.

pkg/rehearse/jobs.go

pkg/rehearse/jobs.go
droslean: addressed!

pkg/rehearse/jobs.go
petr-muller: What I actually had in mind (but haven't described well, re-reading myself) was even more extreme - basically entirely remove this path from `ConfigureRehearsalJobs`. Call it from `main()`. That way `CRJ` can be kept simpler - no need to add new param to it, less responsibilities overall...

But I guess this is fine too.

droslean: It is needed to be here in order to exclude if any changed job is using the specific template with the specific cluster_type. I agree that it would be better to have this outside, but it will need more refactoring. I could open a follow-up for it.

petr-muller: Unless I'm overlooking something, the block [here](https://github.com/openshift/ci-operator-prowgen/pull/122/files#diff-fe857c8e1c6720bb366e42d076a9b101R228) simply adds more rehearsals, where's the exclusion happening?

pkg/rehearse/jobs.go
```diff
@@ -219,9 +225,116 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 		}
 	}
 
+	if allowVolumes {
+		// for the remaining cluster types of each template, pick a `random` corresponding job to rehearse.
+		rehearsals = append(rehearsals, addRandomJobsForChangedTemplates(templates, masterConfigPresubmits, loggers, prNumber)...)
```
petr-muller: Reading more into this, I do not understand why we are using `masterConfigPresubmits` here? That config is basically stale and can contain jobs that were removed or changed in the PR, there's no point in executing these... Why are we not using the PR version of the config?

pkg/rehearse/jobs.go
droslean: oh, Jesus, I removed it when splitting the logic. Nice catch. I will also split the logic to `main` as you suggested.
In the meanwhile, check line 222

pkg/rehearse/jobs.go
droslean: The job that we will pick up, is not supposed to be changed.  At the beginning I was using the PR's config, then I changed it. However, I didn't think the case of the job will be deleted. I will change that.

pkg/rehearse/jobs.go

pkg/rehearse/jobs.go
bbguimaraes: Cluster types are not in `prowgen` for a reason. We don't want to have to update it whenever that list changes. We should get the list from the jobs if it is necessary.

pkg/rehearse/jobs.go
```diff
@@ -222,6 +228,107 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+	templateKey string
+}
+
+func getVolumeIndexAndTempCMName(volumeMounts []v1.VolumeMount, volumes []v1.Volume, templates config.CiTemplates) (*volumeRef, error) {
+	exists, index, templateKey := hasChangedTemplateVolume(volumeMounts, volumes, templates)
+	if !exists {
+		return nil, nil
+	}
+
+	templateData, err := config.GetTemplateData(templates[templateKey])
```
bbguimaraes: We decode the same templates (in `config.GetTemplateData`) countless times. We should store the value when they are first loaded. Many of the new "error handling" `if`s we have are because of that function, even though it is guaranteed not to fail if it succeeded once.

pkg/rehearse/jobs.go
```diff
@@ -222,6 +228,107 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
 	return rehearsals
 }
 
+// AddRandomJobsForChangedTemplates finds random jobs from the PR config that are using a specific template with a specific cluster type.
+// So if a template will be changed, find the jobs that are using a template in combination with the `aws`,`openstack`,`gcs` and `libvirt` cluster types.
+func AddRandomJobsForChangedTemplates(templatesWithClusterTypes map[string]sets.String, templates config.CiTemplates, prConfigPresubmits map[string][]prowconfig.Presubmit, loggers Loggers, prNumber int) config.Presubmits {
```
bbguimaraes: Maybe I am missing something, but how is this random?

pkg/rehearse/jobs.go
```diff
@@ -222,6 +228,107 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get template's data: %s: %v", templates[templateKey], err)
+	}
+	return &volumeRef{
+		index:       index,
+		name:        config.GetTempCMName(config.GetTemplateName(templateKey), templateKey, templateData),
+		templateKey: templateKey}, nil
+}
+
+func getClusterTypeEnvIndex(envVars []v1.EnvVar) (bool, int) {
```
bbguimaraes: Return -1 (as the stdlib does) as the index instead of making the return value `bool, int`.

pkg/rehearse/jobs.go
```diff
@@ -222,6 +228,107 @@ func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.
...
+func hasTemplateFile(job prowconfig.Presubmit, templateFile string) bool {
+	if job.Spec.Containers[0].VolumeMounts != nil {
+		for _, volumeMount := range job.Spec.Containers[0].VolumeMounts {
+			if volumeMount.SubPath == templateFile {
+				return true
+			}
+		}
+	}
+	return false
+}
```
bbguimaraes, line 312: Can't these two functions be implemented using (or replaced by) `getClusterTypeEnvIndex` and `getVolumeIndexAndTempCMName`?

cmd/pj-rehearse/main_test.go
```diff
@@ -152,15 +156,31 @@ func getRehersalsHelper(logger *logrus.Entry, prNumber int) ([]*prowconfig.Presu
...
 	changedCiopConfigs, affectedJobs := diffs.GetChangedCiopConfigs(ciopMasterConfig, ciopPrConfig, logger)
 	changedPresubmits.AddAll(diffs.GetPresubmitsForCiopConfigs(prowPRConfig, changedCiopConfigs, logger, affectedJobs))
 
-	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, ciopPrConfig, prNumber, rehearse.Loggers{Job: logger, Debug: logger}, false, nil)
+	changedTemplates := diffs.GetChangedTemplates(masterTemplates, candidateTemplates, logger)
+	templatesWithClusterTypes := rehearse.GetTemplatesWithClusterTypes(changedTemplates)
+	presubmitsWithChangedTemplates := rehearse.AddRandomJobsForChangedTemplates(templatesWithClusterTypes, changedTemplates, prowPRConfig.JobConfig.Presubmits, loggers, prNumber)
+	changedPresubmits.AddAll(presubmitsWithChangedTemplates)
+
+	rehearsals := rehearse.ConfigureRehearsalJobs(changedPresubmits, ciopPrConfig, prNumber, loggers, true, changedTemplates)
```
bbguimaraes: It has been mentioned before, but the fact that this is a copy of `main.go` is really bad for maintenance. Can we at least call that code from here?

bbguimaraes (changes requested):
I think we can do better than random if we analyze the usage of templates, but this is OK as a first step. I left some comments on specific implementation details. Two suggestions about things that made the review more difficult than it could have been:

- Don't mix refactoring/fixes with features, specially in the same commit. Put them in separate commits, ideally at the beginning of the history (or in a separate PR) if they don't depend on the new code.
- Overall, we have been adding a lot of ad-hoc code that is very similar to existing code. We should try to consolidate new and existing code, refactoring as needed.

pkg/rehearse/jobs.go
droslean: golang maps :) The order is random

pkg/rehearse/jobs.go
droslean: nice idea.

pkg/rehearse/jobs.go
bbguimaraes: It is [unspecified](https://golang.org/ref/spec#For_range), not random. The (current) implementation is pseudo-random with the exact intent of making people not rely on the iteration order. We should use a proper randomization method if that is what we need.

droslean:
@openshift/developer-productivity-test-platform I would like to squash some commits before merging.

pkg/rehearse/jobs.go
droslean: Well, yea I guess we can call it `pseudo-random` better.

pkg/rehearse/jobs.go
droslean: that's sound like a good follow-up

pkg/rehearse/jobs.go
petr-muller: pseudo-random is something [else](https://en.wikipedia.org/wiki/Pseudorandomness) already, this is just unspecified :D

droslean: comment added.

bbguimaraes: There are two issues here that should not be mixed:
- The *current implementation* is not random — we are dealing with computers, so obviously we are talking about pseudo-randomness ("pseudo-pseudo-random" sounded too ridiculous, maybe I should have called it "bad pseudo-randomness"). This is a much smaller problem.
- The *iteration order* is unspecified. This is my biggest issue with this code. Again, if we want (pseudo-)randomness, we should use a proper method. To me, relying on it is fundamentally wrong and deeply ironic, given that whatever "randomness" is used in the iteration order was added to discourage this.

bbguimaraes:
Latest changes LGTM. I still think we are adding a lot more tech debt here than necessary, but am OK with addressing it in subsequent PRs.

pkg/rehearse/jobs.go
droslean: Well, we don't really care if we will pick the same job. Since the map keys are not sorted, we will end up to pick a job from a different repo. At the end, we just want to rehearse one of the jobs that using the specific template and cluster_type

petr-muller: > Again, if we want (pseudo-)randomness

@bbguimaraes like mentioned on slack, I don't think we even need pseudo-randomness. It just gives you some statistic properties about the distribution, but we don't care about that, we just want to avoid hitting the same job every time, for which this is IMHO sufficient.

cmd/pj-rehearse/main_test.go
droslean: I really think that we can drop all of the `main_test.go` code. The integration tests that we have are sufficient for this purpose.
@petr-muller what do you think? Can we kill that code?

pkg/rehearse/jobs.go
droslean: I don't think so. In `hasTemplateFile()` we check a specific template file.

cmd/pj-rehearse/main_test.go
petr-muller: > I really think that we can drop all of the main_test.go code. The integration tests that we have are sufficient for this purpose.

SGTM

droslean: addressed [here](https://github.com/openshift/ci-operator-prowgen/pull/125)

pkg/rehearse/jobs.go
bbguimaraes: You are both justifying the first issue, which I already said is not nearly as important as the second.

I don't see how I can move this argument forward with more comments, so — even though I still think this is wrong — I'll say: overruled by 2x1.

bbguimaraes:
@openshift/developer-productivity-test-platform: is there anything else that needs to be done here? This looks ready to me and I'd like to get the fixes to the integration tests merged ASAP.

petr-muller:
> is there anything else that needs to be done here? This looks ready to me and I'd like to get the fixes to the integration tests merged ASAP.

@bbguimaraes I wanted to have a final look today afternoon, but if it blocks you feel free to merge it

petr-muller (approved):
/lgtm

I must admit I was a bit hesitant to return to this one because the previous versions were a bit harder to review, but I really like how this one unfolded, I like the current version very much!

PR #125, opened 2019-03-29
remove main tests

droslean:
Integration tests that we have already are sufficient for this purpose.


petr-muller (commented):
/lgtm
/hold

LGTM but let's give our dear test obsessed team lead a chance to weigh in

PR #126, opened 2019-03-29
Add a tool for scraping rehearsal metrics from GCS

petr-muller:
Still a bit rough, but should be reviewable mostly for high-level feedback. If the overall structure is OK, I'll polish it. There's a little caching there but the efficiency is not stellar (still need to iterate over a lot of things), but I did not want to make it more complex, because the performance is not that bad (with 20 workers it takes about 40s to run).

### DIFF
--- a/cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
+++ b/cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
@@ -1,0 +1,235 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+
+	"cloud.google.com/go/storage"
+	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
+	"google.golang.org/api/iterator"
+)
+
+const (
+	originCiBucket   = "origin-ci-test"
+	rehearseJobName  = "pull-ci-openshift-release-master-pj-rehearse"
+	metricsPathInRun = "artifacts/rehearse-metrics.json"
+	workersCount     = 20
+)
+
+type scraper struct {
+	cacheDir string
+	bucket   *storage.BucketHandle
+	ctx      context.Context
+}
+
+func scrape(reader io.Reader, base string, run *runInfo) {
+	data, err := ioutil.ReadAll(reader)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to read rehearse metrics")
+		return
+	}
+
+	if err := os.MkdirAll(run.localPrDir(base), 0755); err != nil {
+		logrus.WithError(err).Error("Failed to create a directory to store metrics for PR")
+		return
+	}
+
+	if err := ioutil.WriteFile(run.localMetricsArtifact(base), data, 0644); err != nil {
+		logrus.WithError(err).Error("Failed to save metrics to a local file")
+		return
+	}
+}
+
+type runInfo struct {
+	gcsPrDir  string
+	gcsRunDir string
+}
+
+func (r *runInfo) gcsJobDir() string {
+	jobPrefix := path.Join(r.gcsPrDir, rehearseJobName)
+	return fmt.Sprintf("%s/", jobPrefix) // Needed because path.Join() removes trailing slash
+}
+
+func (r *runInfo) gcsMetricsArtifact() string {
+	return path.Join(r.gcsRunDir, metricsPathInRun)
+}
+
+func (r *runInfo) prNumber() string {
+	return path.Base(r.gcsPrDir)
+}
+
+func (r *runInfo) localPrDir(base string) string {
+	return filepath.Join(base, r.prNumber())
+}
+
+func (r *runInfo) runNumber() string {
+	return path.Base(r.gcsRunDir)
+}
+
+func (r *runInfo) localMetricsArtifact(base string) string {
+	return filepath.Join(r.localPrDir(base), r.runNumber())
+}
+
+// Given a path to a directory storing logs of single run of a rehearsal job for a given PR, scrape
+// the metrics artifact from GCS
+func (s *scraper) scrapeRun(run *runInfo) {
+	// If the metrics are already scraped, do nothing
+	if _, err := os.Stat(run.localMetricsArtifact(s.cacheDir)); err == nil {
+		return
+	}
+
+	reader, err := s.bucket.Object(run.gcsMetricsArtifact()).NewReader(s.ctx)
+	if err != nil {
+		return
+	}
+	defer func() {
+		if err := reader.Close(); err != nil {
+			logrus.WithError(err).Error("Failed to close GCS object reader")
+		}
+	}()
+
+	scrape(reader, s.cacheDir, run)
+}
+
+// Given a path to a directory for a single PR, find all runs of the rehearsal job and
+// scrape the metrics artifact from it
+func (s *scraper) scrapePr(prefix string) {
+	run := &runInfo{gcsPrDir: prefix}
+
+	// Iterate over runs of rehearsal job for a single PR
+	// Example: "pr-logs/pull/openshift_release/3000/pull-ci-openshift-release-master-pj-rehearse/123
+	runs := s.bucket.Objects(s.ctx, &storage.Query{Prefix: run.gcsJobDir(), Delimiter: "/"})
+	for {
+		attrs, err := runs.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			logrus.WithError(err).Error("Failed while iterating over rehearse job run paths using a bucket client")
+			break
+		}
+
+		// Non-empty `Name` attribute means we got a full object name (a file), not a prefix (a dir)
+		// We only care about dirs, not files.
+		if attrs.Name != "" {
+			continue
+		}
+		run.gcsRunDir = attrs.Prefix
+		s.scrapeRun(run)
+	}
+}
+
+type options struct {
+	cacheDir string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.cacheDir, "cache-dir", "", "Path to a directory where scraped metrics data will be saved. If not provided, cache will not be used.")
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("Failed to parse options")
+	}
+	return o
+}
+
+func main() {
+	o := gatherOptions()
+
+	ctx := context.Background()
+	client, err := storage.NewClient(ctx)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to create client")
+	}
+
+	if o.cacheDir == "" {
+		o.cacheDir, err = ioutil.TempDir("", "")
+		if err != nil {
+			logrus.WithError(err).Fatal("Failed to create a temporary directory")
+		}
+		defer func() {
+			if err := os.RemoveAll(o.cacheDir); err != nil {
+				logrus.Warnf("Failed to clean up a temporary directory %s (err=%v)", o.cacheDir, err)
+			}
+		}()
+	}
+
+	bucket := client.Bucket(originCiBucket)
+
+	// We will iterate over directories under openshift_release, each corresponds to a single PR to openshift/release
+	// Example: "pr-logs/pull/openshift_release/3000
+	pullRequests := bucket.Objects(ctx, &storage.Query{
+		Prefix:    "pr-logs/pull/openshift_release/",
+		Delimiter: "/",
+	})
+
+	scraper := scraper{ctx: ctx, bucket: bucket, cacheDir: o.cacheDir}
+	prPrefixes := make(chan string, 100)
+	progress := make(chan string, 100)
+
+	sem := semaphore.NewWeighted(workersCount)
+
+	// Spawn X workers. The workers read PR directory paths from a channel, scrape all runs for a given PR and feed
+	// the completed paths to another channel. The workers finish once the input channel is closed.
+	for w := 0; w < workersCount; w++ {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			logrus.WithError(err).Error("Failed to acquire semaphore to spawn a worker")
+			break
+		}
+		go func() {
+			defer sem.Release(1)
+			for prefix := range prPrefixes {
+				scraper.scrapePr(prefix)
+				progress <- prefix
+			}
+		}()
+	}
+
+	// Iterate over PR directories and send them to the workers for scraping.
+	go func() {
+		defer close(progress)
+		for {
+			attrs, err := pullRequests.Next()
+			if err == iterator.Done {
+				break
+			}
+			if err != nil {
+				logrus.WithError(err).Error("Failed while iterating over PR directory paths using a bucket client")
+				break
+			}
+
+			// Non-empty `Name` attribute means we got a full object name (a file), not a prefix (a dir)
+			// We only care about dirs, not files.
+			if attrs.Name != "" {
+				continue
+			}
+
+			prPrefixes <- attrs.Prefix
+		}
+
+		// Close the channel so that workers stop iterating after processing all directories
+		// Then wait for all workers to finish.
+		close(prPrefixes)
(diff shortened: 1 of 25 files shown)


cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	if _, err := os.Stat(localMetricsPath); err == nil {
+		return
+	}
+
+	reader, err := s.bucket.Object(metricsPath).NewReader(s.ctx)
+	if err != nil {
+		return
+	}
+	data, err := ioutil.ReadAll(reader)
+	_ = reader.Close()
```
stevekuznetsov: is this a typo?

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	if err := ioutil.WriteFile(localMetricsPath, data, 0644); err != nil {
+		fmt.Printf("could not write metrics to %s err=%v\n", localMetricsPath, err)
+		return
+	}
+}
+
+func (s *scraper) scrapePr(prefix string) {
+	prNumber := path.Base(prefix)
+	jobPrefix := path.Join(prefix, rehearseJobName)
+	jobPrefix = fmt.Sprintf("%s/", jobPrefix) // Needed because path.Join() removes trailing slash
```
stevekuznetsov: yeah technically you should not use `path.Join` for URLs but whatever

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+		attrs, err := runs.Next()
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			fmt.Printf("Next() failed err=%v\n", err)
+			break
+		}
+
+		// Leaf (a file), not a prefix (a dir)
```
stevekuznetsov: Could be more clear like

```
// We only care about files, not dirs
```

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	cacheDir string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.cacheDir, "cache-dir", "", "Path to a directory where scraped metrics data will be saved. If not provided, cache will not be used.")
+
+	_ = fs.Parse(os.Args[1:])
```
stevekuznetsov: logrus.fatal on this one

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func main() {
+	o := gatherOptions()
+	var err error
+
+	if o.cacheDir == "" {
+		o.cacheDir, err = ioutil.TempDir("", "")
+		if err != nil {
+			_, _ = fmt.Fprintf(os.Stderr, "Failed to create a temporary directory: %v", err)
```
stevekuznetsov: use `logrus.Fatal.WithError()`

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+			os.Exit(1)
+		}
+		defer os.RemoveAll(o.cacheDir)
```
stevekuznetsov: can this error?

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+		Prefix:    "pr-logs/pull/openshift_release/",
+		Delimiter: "/",
+	})
+
+	scraper := scraper{ctx: ctx, bucket: bucket, cacheDir: o.cacheDir}
+	prPrefixes := make(chan string, 100)
+	progress := make(chan string, 100)
+
+	wg := sync.WaitGroup{}
+	wg.Add(20)
```
stevekuznetsov: do you want this to be configurable?

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+
+	for w := 0; w < 20; w++ {
```
stevekuznetsov: Can you use a semaphore instead of this approach? This is the older approach before semaphores were added to the language

stevekuznetsov (commented):
approach LGTM

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
droslean: You can `url.Parse` them and use the `ResolveReference`.
```go
u, _ := url.Parse(prefix)
relative, _ := url.Parse(rehearseJobName)
jobPrefix := u.ResolveReference(relative)
```

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+		go func() {
+			for prefix := range prPrefixes {
+				scraper.scrapePr(prefix)
+				progress <- prefix
+			}
+			wg.Done()
```
droslean: this should be in `defer`

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+			}
+
+			if attrs.Name != "" {
+				continue
+			}
+
+			prPrefixes <- attrs.Prefix
+		}
+		close(prPrefixes)
+		wg.Wait()
```
droslean: its better to close the channels out of the goroutine.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+	progress := make(chan string, 100)
+
+	wg := sync.WaitGroup{}
+	wg.Add(20)
+
+	for w := 0; w < 20; w++ {
+		go func() {
+			for prefix := range prPrefixes {
+				scraper.scrapePr(prefix)
+				progress <- prefix
```
droslean, line 192: this should be like:
```go
defer wg.Done()
for {
    select {
        prefix: = < -prPrefixes
        scraper.scrapePr(prefix)
        progress < -prefix
        case <-done:
            close(prPrefixes)
            close(progress)
            return
    }
}
}
```

droslean: you need also the other channel:     done := make(chan struct{})

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+				break
+			}
+
+			if attrs.Name != "" {
+				continue
+			}
+
+			prPrefixes <- attrs.Prefix
+		}
+		close(prPrefixes)
```
droslean, line 221: Here you can send to `done` channel.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,172 @@
...
+		wg.Wait()
+		close(progress)
+	}()
+
+	counter := 0
+	for done := range progress {
```
droslean, line 229: its a go idiom to use the `for{ select {} }` to handle also other signal channels to gracefully terminate.  
Also you can check if the channel closed. Like:
```go
 	for done, ok := range progress {
if !ok {
// we are done
os.Exit(0) 
}
}
```

droslean (commented):
In general, this is a good start. I left some comments regarding the channels.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: Not really, GoLand suggested this as an explicit error ignore, so I used it. I'll probably just defer it.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: I think I'll go with the `path.Join` way, the object names behave more like files than URLs, so `path` operations are probably a bit more readable, too. The only problem is that the separator is a part of the object name, not an actual separator.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: I guess. I wanted to avoid talking about files and dirs because they are not actually files and dirs, but meh.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
stevekuznetsov: Yeah i think I usually do

```go
defer func(reader io.Reader) {
    if err := reader.Close(); err != nil {
        logrus.WithError(err).Fatal("whatever")
    }
}(reader)
```

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
stevekuznetsov: I guess I meant this is a confusing comment so making it more clear with whatever appropriate words would be good

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: not really, but I'll extract it to a named constant at least

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: Sure.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: Yeah, will defer a func that will at least log the failure.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: About the "check if the channel is closed" part, are you sure? The [spec](https://golang.org/ref/spec#For_range) says _If the range expression is a channel, at most one iteration variable is permitted_. It also does not make sense, I understand that when the channel is closed and all values are exhausted, the iteration simply ends...

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: Why _should_ this be like that? What's wrong? To me it looks more cluttered and needs an additional channel just to communicate the same thing that's communicated here by simply closing the channel.

petr-muller:
I addressed the straightforward feedback and will polish the code a bit more.

@droslean Can you elaborate a bit further on how do you suggest to change the channel code? I'll need help understanding how the code could be improved, in particular:

Why is the `for { select {} }` pattern preferable to the simple `for value := channel` iteration? The simple iteration looks more readable to me and I haven't found any documentation that would discourage its use. You say _... idiom to use the for{ select {} } to handle also other signal channels to gracefully terminate_, but here's just one input channel, not more, which I understand is not the use case for `select`...

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,185 @@
...
+
+	for w := 0; w < workersCount; w++ {
+		if err := sem.Acquire(ctx, 1); err != nil {
+			logrus.WithError(err).Error("Failed to acquire semaphore to spawn a worker")
+			break
+		}
+		go func() {
+			defer sem.Release(1)
+			for prefix := range prPrefixes {
+				scraper.scrapePr(prefix)
```
stevekuznetsov, line 191: Err -- I think you need to limit concurrency _here_:

```go
sem := semaphore.NewWeighted(workersCount)

for prPrefix := range prPrefixes {
    go func(prefix string) {
		if err := sem.Acquire(ctx, 1); err != nil {
			logrus.WithError(err).Error("Failed to acquire semaphore to spawn a worker")
			prPrefixes<-prefix
			return
		}
		defer sem.Release(1)
		// work on the prefix
    }(prPrefix)
}
```

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
stevekuznetsov: Yeah unless you want to cancel in the middle you don't need the `range` and this will stop looping when the channel is closed

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,185 @@
...
+			if err != nil {
+				fmt.Printf("Next() failed: err=%v\n", err)
+				break
+			}
+
+			if attrs.Name != "" {
+				continue
+			}
+
+			prPrefixes <- attrs.Prefix
```
stevekuznetsov, line 216: You could also just remove the need for `prPrefixes`, run the content of this goroutine in the main thread and attempt to spawn a worker here -- we would read new prefixes as we need to instead of upfront but that's not a huge hit in cost, right?

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
```diff
@@ -0,0 +1,185 @@
...
+		}
+		close(prPrefixes)
+		if err := sem.Acquire(ctx, workersCount); err != nil {
+			logrus.WithError(err).Error("Failed to acquire semaphore to wait on workers to finish")
+		}
+	}()
+
+	counter := 0
+	for done := range progress {
```
stevekuznetsov, line 229: This is a bit clumsy, you could use a locked value or atomic updates instead

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: I don't think so, acquiring in the spawned goroutine would create the (probably rare but still) opportunity to hit the final L173 `Acquire`  first, deadlocking the workers.

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: You mean for the counter? Isn't it an overkill (locked value certainly is...) for what is basically a activity indicator?

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
petr-muller: You mean to spawn short-lived goroutines to scrape just a single PR? It's possible, but I'm not sure it's better than the long-lived workers over the channel (not worse either - there's no per-worker expensive setup, so short-lived goroutines would be fine). 

I do not understand the comment about reading prefixes upfront, though - reading prefixes would still need to be implemented via looping over the `pullRequests` iterator, no? This goroutine reads prefixes and sends them to the channel as fast as the workers pop them, where's the upfront read?

petr-muller:
Removed WIP: I polished the code a bit, added few comments and extracted the paths-manipulating code to a separate struct, added tests for it. I haven't changed the concurrency structure, because I think it's correct and I prefer it to suggested alternatives (I answered individual suggestions inline).

cmd/pj-rehearse-metrics/pj-rehearse-metrics.go
stevekuznetsov: The approach right now reads them all upfront and sends to the channel, the channel holds all of the values (sans whichever have started to be processed). Doing it just-in-time will read as we process.

stevekuznetsov (commented):
The whole point of the weighted semaphore is to not have the explicit pool -- what you're doing doesn't need the semaphore at all, you're basically just using it as a WaitGroup. The concurrency-bounded action is operating on the prefix, so we should acquire and release around that code. if every thread only ever acquires once then there's no real need for the semaphore, WaitGroup is sufficient. But, I suggested that approach in review 

/shrug
:man_facepalming:

PR #127, opened 2019-04-03
rehearse: use updateconfig to generate CMs

bbguimaraes:
I am still working on templates, as that will require more fundamental changes, but the basic mechanism is already reviewable.

### DIFF
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -20,6 +20,11 @@ import (
 	kutilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	prowgithub "k8s.io/test-infra/prow/github"
+	_ "k8s.io/test-infra/prow/hook"
+	prowplugins "k8s.io/test-infra/prow/plugins"
+	"k8s.io/test-infra/prow/plugins/updateconfig"
 )
 
 // CiTemplates is a map of all the changed templates
@@ -58,17 +63,27 @@ func getTemplates(templatePath string) (CiTemplates, error) {
 
 // TemplateCMManager holds the details needed for the configmap controller
 type TemplateCMManager struct {
-	cmclient corev1.ConfigMapInterface
-	prNumber int
-	logger   *logrus.Entry
+	cmclient         corev1.ConfigMapInterface
+	configUpdaterCfg prowplugins.ConfigUpdater
+	prNumber         int
+	releaseRepoPath  string
+	logger           *logrus.Entry
 }
 
 // NewTemplateCMManager creates a new TemplateCMManager
-func NewTemplateCMManager(cmclient corev1.ConfigMapInterface, prNumber int, logger *logrus.Entry) *TemplateCMManager {
+func NewTemplateCMManager(
+	cmclient corev1.ConfigMapInterface,
+	configUpdaterCfg prowplugins.ConfigUpdater,
+	prNumber int,
+	releaseRepoPath string,
+	logger *logrus.Entry,
+) *TemplateCMManager {
 	return &TemplateCMManager{
-		cmclient: cmclient,
-		prNumber: prNumber,
-		logger:   logger,
+		cmclient:         cmclient,
+		configUpdaterCfg: configUpdaterCfg,
+		prNumber:         prNumber,
+		releaseRepoPath:  releaseRepoPath,
+		logger:           logger,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


pkg/config/template.go
```diff
@@ -26,6 +26,11 @@ import (
 
 	"k8s.io/client-go/kubernetes/scheme"
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	prowgithub "k8s.io/test-infra/prow/github"
+	_ "k8s.io/test-infra/prow/hook"
```
stevekuznetsov, line 25: I should add a non-validating load so we don't need this -.-

stevekuznetsov (commented):
@bbguimaraes you want to make sure this is ready to go for the profiles and then do templates in another PR?

pkg/config/template.go
bbguimaraes: Does the configuration check we have also validate plugins? If not (and even if it is?), validating this may be an additional benefit of `rehearse`.

bbguimaraes:
I am considering it. The template changes are almost done, they would make the PR significantly larger and are mostly orthogonal.

It would be nice to get #122 in before this so I can add proper integration tests, I'll take a look and see what's left in there.

pkg/config/template.go
stevekuznetsov: It does, yeah

petr-muller:
LGTM, nice! Would prefer to merge this and have another PR for templates...

bbguimaraes:
Rebased and updated. If you've reviewed this before, I tried to keep the diff to a minimum: check only the last commit. I'll squash when approved. This can be merged as is, but you can also check the template PR:

https://github.com/openshift/ci-operator-prowgen/pull/146

Main differences from the previous version:
- fixed path in `main.go` (caught by https://github.com/openshift/ci-operator-prowgen/pull/140)
- fixed to create only necessary profiles (caught by https://github.com/openshift/ci-operator-prowgen/pull/143)
- added necessary files for the new integration tests in https://github.com/openshift/ci-operator-prowgen/pull/140
- added an `update` reactor to the dry-run client, the `create` reactor is kept because of the old template code and is removed in the next PR

pkg/config/template.go
```diff
@@ -103,43 +108,53 @@ func (c *TemplateCMManager) CreateCMTemplates(templates CiTemplates) error {
...
+			if err != nil || info.IsDir() {
+				return err
+			}
+			if path, err = filepath.Rel(dir, path); err == nil {
+				changes = append(changes, prowgithub.PullRequestChange{
+					Filename: path,
+					Status:   string(prowgithub.PullRequestFileModified),
+				})
+			}
 			return err
```
stevekuznetsov, line 143: It's impossible for `filepath.Rel` to error here right? Might be worth a comment

pkg/config/template.go
```diff
@@ -103,43 +108,53 @@ func (c *TemplateCMManager) CreateCMTemplates(templates CiTemplates) error {
...
+	var errs []error
+	for cm, data := range updateconfig.FilterChanges(cms, changes, c.logger) {
+		profile := strings.TrimPrefix(cm.Name, "cluster-profile-")
+		for _, p := range profiles {
+			if p.Name == profile {
+				profile = p.CMName()
+				break
+			}
+		}
+		err := c.createCM(&v1.ConfigMap{
```
stevekuznetsov, line 158: Update should do creation, why do we need this?

bbguimaraes: To add the labels. This is something we could easily add to the upstream library in the future.

pkg/config/template.go
bbguimaraes: Added.

bbguimaraes:
Incorporated @petr-muller's [suggestion](https://github.com/openshift/ci-operator-prowgen/pull/146#discussion_r276610780) and added an extra test fix that got lost during the refactorings.

pkg/config/template.go
stevekuznetsov: Can we rely on upstream to create and just add labels ourselves after?

bbguimaraes: That was the other option, but it would require a `Get` + `Update` here (`updateconfig.Update` doesn't return what it created), while `updateconfig` already does a `Get` first.

bbguimaraes: For reference: https://github.com/kubernetes/test-infra/blob/50de43197c2fdf95570eb414e3510a0b562704f9/prow/plugins/updateconfig/updateconfig.go#L90

bbguimaraes: Doing that also leaves a (very unlikely, but still…) window where we could create CMs that would never be removed if we failed before labeling them.

bbguimaraes:
Rebase and squashed.

Some small changes were necessary because of an upstream change to the interface of the `updateconfig` functions: https://github.com/kubernetes/test-infra/commit/8ca35465a98da0c769756bc75e9e00fdc851001d.

stevekuznetsov (approved):
/lgtm
/approve

PR #129, opened 2019-04-05
Allow config-brancher to mirror promotion outputs

stevekuznetsov:
When we are creating a release branch for the same version that the
development branch is using, that release branch needs to mirror the
promotion config from the development branch but in a disabled manner.
The config-brancher tool can now create these mirrors and unmirror to do
the actual bump for the development branch in the future.

/assign @smarterclayton 


stevekuznetsov:
Going to iterate after this but want a new image built

PR #133, opened 2019-04-10
replace git-ref arg with extra_refs

droslean:
Replace the `--git-ref` argument with `extra_refs` in rehearsals. 
That will allow `ci-operator` to clone the main repository that the test will run against and keep the existing `JOB_SPEC` that points to the PR in `release` repository.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -85,10 +85,16 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber 
 	shortName := strings.TrimPrefix(source.Context, "ci/prow/")
 	rehearsal.Context = fmt.Sprintf("ci/rehearse/%s/%s/%s", repo, branch, shortName)
 	rehearsal.RerunCommand = defaultRehearsalRerunCommand
-
-	gitrefArg := fmt.Sprintf("--git-ref=%s@%s", repo, branch)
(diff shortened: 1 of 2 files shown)


petr-muller:
@droslean So I tested this on https://github.com/openshift/release/pull/3431 and it does not seem to work, the target repo bits are pulled to a [different location](https://openshift-gce-devel.appspot.com/build/origin-ci-test/pr-logs/pull/openshift_release/3431/rehearse-3431-pull-ci-openshift-ci-operator-prowgen-master-moar-lint/2?log#log) than the builds expect them... 

Does this need some fix also in the ci-operator?

stevekuznetsov:
Working directory for the test will remain from `refs`, not `extra_refs`

petr-muller:
> Working directory for the test will remain from `refs`, not `extra_refs`

We could work around that. `ci-operator` can get a `-C`-like param to change its working dir, we may do some hack in the rehearse itself (change cwd before calling ci-operator), or maybe `ProwJobSpec` could get an option to make cwd for the test configurable.

droslean:
I am ok with just hacking the rehearse tool. Maybe changing the command of the job to chdir first.

petr-muller:
Personally I like the configurable CWD in `ProwJobSpec` the most. It's something that could be useful together with `extra_refs` also in other cases.

droslean:
/close

No `extra_refs` are needed. The `authors` problems, fixed in [ci-operator/pull/321](https://github.com/openshift/ci-operator/pull/321/)

PR #136, opened 2019-04-11
rehearse: check templates in integration tests

bbguimaraes:


droslean (commented):
That looks good. I wondering if its better to add a PretendFunc to the fake cm client to print the cms as yaml, instead of passing the `o.dryRun` all around?

PR #137, opened 2019-04-11
Add a tool to publish merge blocking issues for releases

stevekuznetsov:
When we are automatically fast-forwarding branches from the development
branch for future releases, it is not valid to merge to those branches
directly. The new `blocking-issue-creator` issue posts issues to the
repositories for which we are managing future release branches in the
format that `tide` expects to block automatic merging to those branches.

/assign @smarterclayton 
/cc @cjwagner

### DIFF
--- a/cmd/blocking-issue-creator/main.go
+++ b/cmd/blocking-issue-creator/main.go
@@ -1,0 +1,247 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/openshift/ci-operator/pkg/api"
+	githubql "github.com/shurcooL/githubv4"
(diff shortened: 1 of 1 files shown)


smarterclayton:
looked mostly correct to me

petr-muller (approved):
/lgtm

Reviewed and LGTM. Would be a little more readable if some fragments would be extracted to methods, but it's not that long and the logic is clear so /shrug.

cjwagner:
Is there some benefit to having a Tide query include branches that will always be forbidden to merge to? I would think that this use case would be best addressed by using `includedBranches` and `excludedBranches` in the queries so that branches that are forbidden to merge to are never considered by Tide. https://github.com/kubernetes/test-infra/blob/a91153f323102e6a008da8b9781ad3bb13c81ee9/prow/config/tide.go#L150-L151

PR #138, opened 2019-04-11
Generate prepublish job in presubmits when config prepublish is set

chancez:
Generates prepublish jobs for https://github.com/openshift/ci-operator/pull/311

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -377,6 +377,9 @@ func generateJobs(
 		if configSpec.PromotionConfiguration != nil {
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *generatePostsubmitForTest("images", info, true, labels, generatePodSpec(info, "[images]", additionalPostsubmitArgs...)))
 		}
+		if configSpec.PrepublishConfiguration != nil {
+			presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest("prepublish", info, generatePodSpec(info, "[prepublish]", additionalPresubmitArgs...)))
+		}
 	}
 
 	return &prowconfig.JobConfig{
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -365,11 +365,11 @@
 
 [[projects]]
   branch = "master"
-  digest = "1:334deadc0e3569d00920be459aa7072288635ed728b70e8f2584e86a1c1dee81"
+  digest = "1:59c4be19224b064610e20830e219ee26e80521f326d5341379126947a1c3670f"
   name = "github.com/openshift/ci-operator"
   packages = ["pkg/api"]
   pruneopts = ""
-  revision = "ccb6f432a6b149c1544c09038868b9a34b7434f7"
+  revision = "6c5bcba27577c260e11ba87d37648d21e033aa3a"
 
 [[projects]]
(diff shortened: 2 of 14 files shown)


stevekuznetsov:
LMAO the bug from the other PR broke images here -- ping me when the smoke clears

cmd/ci-operator-prowgen/main.go
```diff
@@ -377,6 +377,9 @@ func generateJobs(
 		if configSpec.PromotionConfiguration != nil {
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *generatePostsubmitForTest("images", info, true, labels, generatePodSpec(info, "[images]", additionalPostsubmitArgs...)))
 		}
+		if configSpec.PrepublishConfiguration != nil {
+			presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest("prepublish", info, generatePodSpec(info, "[prepublish]", additionalPresubmitArgs...)))
```
petr-muller, line 381: Just a thought. For some reason I was expecting that we'll do the prepublishing within the `images` job, by appending the necessary `[prepublish]` target params.  But a separate job is also possible, I guess.

As a CI system, we don't care, both jobs will race for building the individual images and then the prepublish one will publish them.

I'm wondering about the effect on PR authors, though - having too many jobs may be confusing, especially if multiple ones fail on a single reason. I think separate job is worth having if its unique steps can fail in useful way - I would not expect prepublishing to fail unless we have infrastructure problems, which need admin intervention anyway, so maybe I would prefer to prepublish within the images job.

cmd/ci-operator-prowgen/main.go
```diff
@@ -377,6 +377,9 @@ func generateJobs(
 		if configSpec.PromotionConfiguration != nil {
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *generatePostsubmitForTest("images", info, true, labels, generatePodSpec(info, "[images]", additionalPostsubmitArgs...)))
 		}
+		if configSpec.PrepublishConfiguration != nil {
```
petr-muller, line 380: The condition makes the prepublish run only when there's the top-level `prepublish` config, is that intentional? Nothing would be published if you only configured the individual steps...

cmd/ci-operator-prowgen/main.go
chancez: Ah I didn't know each job could do more than one target. That makes sense. I think you've got a good point that this could easily be part of the images job.

cmd/ci-operator-prowgen/main.go
chancez: Good point, but I'm also realizing, then, how does anyone ever use rawSteps? Im guessing if you use rawSteps that you might also be required to write the prow job manually? I don't see code in prowgen that handles any kind of raw steps right now.

petr-muller:
/hold

Not likely to merge until we deal with ci-operator feature freeze https://github.com/openshift/ci-operator/pull/329#issuecomment-482866451

cmd/ci-operator-prowgen/main.go
stevekuznetsov: Raw steps get used as intermediates for leaf nodes that do get jobs generated for them

chancez: Right. I was asking how do they get run if you directly specify `rawSteps`

PR #139, opened 2019-04-15
rehearse: do not decode templates

bbguimaraes:
Decoding the templates is not necessary for any operation in `rehearse`. Treating the files as an array of bytes is enough to tell if it changed and to populate the temporary `ConfigMap`s.

Not decoding simplifies the code and greatly simplifies the tests.


bbguimaraes:
`OWNERS` in `vendor/`, didn't see that coming.

/uncc danwinship bparees

pkg/config/template.go
```diff
@@ -40,24 +34,16 @@ const (
 )
 
 func getTemplates(templatePath string) (CiTemplates, error) {
-	templates := make(map[string]*templateapi.Template)
+	templates := make(CiTemplates)
 	err := filepath.Walk(templatePath, func(path string, info os.FileInfo, err error) error {
-		if err != nil {
-			return fmt.Errorf("prevent panic by handling failure accessing a path %q: %v", path, err)
+		if err != nil || info.IsDir() || !(strings.HasSuffix(path, ".yaml") || strings.HasSuffix(path, ".yml")) {
```
droslean, line 39: this check can be done with `isYAML()` as well.
```go
if err != nil || isYAML(path, info) {
```

droslean (commented):
nice one.

pkg/config/template.go
bbguimaraes: I inlined that function because it is only ever used here, but that got lost when I moved commits around. Updated.

droslean: why we want `(strings.HasSuffix(path, ".yaml")` to be false?

PR #146, opened 2019-04-17
rehearse: use updateconfig for template CMs

bbguimaraes:
Based on https://github.com/openshift/ci-operator-prowgen/pull/127.

### DIFF
--- a/pkg/config/template.go
+++ b/pkg/config/template.go
@@ -20,6 +20,11 @@ import (
 	kutilerrors "k8s.io/apimachinery/pkg/util/errors"
 
 	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	prowgithub "k8s.io/test-infra/prow/github"
+	_ "k8s.io/test-infra/prow/hook"
+	prowplugins "k8s.io/test-infra/prow/plugins"
+	"k8s.io/test-infra/prow/plugins/updateconfig"
 )
 
 // CiTemplates is a map of all the changed templates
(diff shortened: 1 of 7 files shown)


bbguimaraes:
I found changing the names of the ConfigMaps in a copy of the `updateconfig` spec cleaner than the previous approach of retroactively identifying the source of each CM. I didn't want to add that to #127, it is done here.

The rest of https://github.com/openshift/ci-operator-prowgen/pull/146/commits/3624b0aed6001f1663f3a27fe7ab2d4f26437626 is just the extraction of the profile code into separate functions, https://github.com/openshift/ci-operator-prowgen/pull/146/commits/868e10d19d9921db2bce11e5daa16b21da94e222 changes the template code to use those and changes the tests accordingly.

pkg/config/template.go
```diff
@@ -116,47 +116,84 @@ func (g osFileGetter) GetFile(filename string) ([]byte, error) {
...
+	for cm, data := range updateconfig.FilterChanges(cmSpec, changes, c.logger) {
+		c.logger.WithFields(logrus.Fields{"cm-name": cm.Name}).Info("creating rehearsal configMap")
+		if err := c.createCM(dir, cm.Name, data); err != nil {
 			errs = append(errs, err)
 		}
 	}
 	return kutilerrors.NewAggregate(errs)
 }
 
+func (c *TemplateCMManager) CreateClusterProfiles(dir string, cmSpec map[string]prowplugins.ConfigMapSpec, profiles []ClusterProfile) error {
```
petr-muller: I haven't noticed this in #127, but I would slightly prefer to have the path/config items in the manager struct to keep Create* method API dead simple. I think the struct can keep the "static stuff that we need to create templates/CMs" and Create* methods can just take params what to create.

bbguimaraes: Good call. Both PRs updated.

bbguimaraes:
Run tests, *then* push.

bbguimaraes:
Rebased and squashed.

PR #150, opened 2019-04-23
Tool to update team membership from downloaded spreadsheet

stevekuznetsov:
@openshift/openshift-team-developer-productivity-test-platform

### DIFF
--- a/cmd/team-member-updater/main.go
+++ b/cmd/team-member-updater/main.go
@@ -1,0 +1,283 @@
+package main
+
+import (
+	"encoding/csv"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/github"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/config/org"
+	"sigs.k8s.io/yaml"
+)
+
+type options struct {
+	membersPath string
+	reposPath   string
+	configPath  string
+	logLevel    string
+}
+
+func (o *options) Validate() error {
+	if o.membersPath == "" {
+		return errors.New("required flag --members was unset")
+	}
+
+	if o.reposPath == "" {
+		return errors.New("required flag --repos was unset")
+	}
+
+	if o.configPath == "" {
+		return errors.New("required flag --config was unset")
+	}
+
+	level, err := logrus.ParseLevel(o.logLevel)
+	if err != nil {
+		return fmt.Errorf("invalid --log-level: %v", err)
+	}
+	logrus.SetLevel(level)
+	return nil
+}
+
+func gatherOptions() options {
+	o := options{}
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
@bbguimaraes FYI -- going to try to break out the go modules thing

cmd/team-member-updater/main.go
```diff
@@ -0,0 +1,230 @@
...
+//   8: time zone
+//   9: empty
+//  10: role
+//  11: empty
+//  12: notes
+//  13: empty
+//  14: valid team
+//  15: email
+//  16: github id
+func teamMemberFromRow(row []string) teamMember {
```
droslean, line 253: This assumes/trust that the `row` will be always a specific format. Can we validate this here and return an error before trying to access a nil index?

cmd/team-member-updater/main.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	githubIDsbyGroup := map[string]sets.String{}
+	githubIDsbyTeam := map[string]sets.String{}
+	githubIDsbyRole := map[string]sets.String{}
+	allGithubIDs := sets.NewString(orgConfig.Orgs["openshift"].Members...)
+	for _, member := range members {
+		allGithubIDs.Insert(member.githubID)
+		if member.group != "" {
+			if githubIDsbyGroup[member.group] == nil {
```
droslean: IMHO since this check is repeating also in other parts of the code, do you think that its better to have
```go
type githubIDs struct {
    byGroup map[string]sets.String
    byTeam  map[string]sets.String
    byRole  map[string]sets.String
}
```
and construct them?

or spliting the logic with a helper function?
```go
func populateGithubIDs(key,value string, set map[string]sets.String ) {
    if set[key] == nil {
        set[key] = sets.NewString(value)
    } else {
        set[key].Insert(value)
    }
}
```

stevekuznetsov:
@openshift/openshift-team-developer-productivity-platform tool is ready, creates teams and assigns them repository roles

/cc @coverprice

cmd/team-member-updater/main.go
```diff
@@ -0,0 +1,283 @@
...
+func main() {
+	o := gatherOptions()
+	if err := o.Validate(); err != nil {
+		logrus.Fatalf("Invalid options: %v", err)
+	}
+
+	var orgConfig struct {
+		Orgs map[string]org.Config `json:"orgs,omitempty"`
+	}
+	rawConfig, err := ioutil.ReadFile(o.configPath)
```
bbguimaraes, line 70: As someone who doesn't initially know how the program should operate, it's hard to keep track of this number of local variables. E.g. none of the `raw*` or `*Records` variable needs to exist a few lines after it is created. Can we break `main` into a few more tightly-scoped functions?

bbguimaraes (commented):
One structural suggestion, the rest LGTM.

stevekuznetsov:
I think @coverprice wants this in Python anyway

/close

coverprice:
Yeah we'll rewrite this in Python, and eventually migrate the data to LDAP or Atlas so that it's easier to manage.

PR #151, opened 2019-04-24
add vsphere cluster type

droslean:
Add `vsphere` cluster type when picking new jobs to rehearse a single template change.


petr-muller (approved):
/lgtm
/approve

PR #152, opened 2019-04-24
Use go modules

stevekuznetsov:
I have no idea why the vendor diff is as huge as it is, but this updates us to use `go mod` and bumps to latest test-infra.

@openshift/openshift-team-developer-productivity-test-platform


petr-muller (approved):
/lgtm

Builds, passes tests and there does not seem to be anything suspicious

PR #154, opened 2019-04-30
Update references to input images on bump

stevekuznetsov:
When we change the tag specificatio and promotion we also need to chase
those references throughout the config for any other inputs that users
have set.

/assign @smarterclayton 
/cc @openshift/openshift-team-developer-productivity-platform

### DIFF
--- a/pkg/promotion/promotion.go
+++ b/pkg/promotion/promotion.go
@@ -4,12 +4,11 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"regexp"
-
 	cioperatorapi "github.com/openshift/ci-operator/pkg/api"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/flagutil"
+	"regexp"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/promotion/promotion.go
```diff
@@ -34,7 +33,12 @@ func isDisabled(configSpec *cioperatorapi.ReleaseBuildConfiguration) bool {
 func buildOfficialImages(configSpec *cioperatorapi.ReleaseBuildConfiguration) bool {
 	promotionNamespace := extractPromotionNamespace(configSpec)
 	promotionName := extractPromotionName(configSpec)
-	return (promotionNamespace == okdPromotionNamespace && promotionName == okd40Imagestream) || promotionNamespace == ocpPromotionNamespace
+	return RefersToOfficialImage(promotionName, promotionNamespace)
+}
+
+// RefersToOfficialImage determines if an image is official
+func RefersToOfficialImage(name, namespace string) bool {
+	return (namespace == okdPromotionNamespace && name == okd40Imagestream) || namespace == ocpPromotionNamespace
```
smarterclayton, line 41: This can just be the “origin” namespace now

pkg/promotion/promotion.go
```diff
@@ -53,6 +57,11 @@ func extractPromotionName(configSpec *cioperatorapi.ReleaseBuildConfiguration) s
 	return ""
 }
 
+// IsBumpable determines if the dev branch should be bumped or not
+func IsBumpable(branch, currentRelease string) bool {
+	return branch != fmt.Sprintf("openshift-%s", currentRelease)
```
smarterclayton, line 62: This is confusing

stevekuznetsov: We don't bump dev branches that are release branches and we operate on one branch file at a time so it's impossible to say otherwise

pkg/promotion/promotion.go
stevekuznetsov: any stream in origin namespace? let's change this in a follow-up

pkg/promotion/promotion.go
smarterclayton: ok

PR #156, opened 2019-05-03
rehearse: use full paths, modify tests

bbguimaraes:
Small fixes to make the test suite more realistic so we catch errors like https://github.com/openshift/ci-operator-prowgen/pull/146. Clean up some inconsistencies.

Each commit is a fairly small and independent change. Tests pass throughout.


petr-muller (approved):
/lgtm

PR #158, opened 2019-05-08
add gcp cluster type for selection

droslean:


petr-muller (approved):
why not I guess :D
/lgtm

droslean:
> why not I guess :D

@petr-muller related PR that the template was changed but we didn't trigger the rehearse:
https://github.com/openshift/release/pull/3691

I triggered it manually. `ci/rehearse/openshift/jenkins/openshift-3.11/e2e-gcp `

stevekuznetsov:
Don't we have constants for these somewhere? Ew. We need to re-think what is going on with these constants. Is it better when we move all templates to Go code?

bbguimaraes:
I mentioned it in the [original review](https://github.com/openshift/ci-operator-prowgen/pull/122#discussion_r269568616): we didn't have these constants in `ci-operator-prowgen` to avoid this. The ones we use to generate jobs are in `ci-operator`, but even there we don't have a hard-coded list that has to be manually updated.

At this point in the execution in `rehearse`, we have all jobs from the master and candidate branches loaded in memory, why not derive the list from them?

PR #159, opened 2019-05-09
rehearse: use updateconfig for template CMs

bbguimaraes:
Take two.

94666d37 is new. I am not sure if it is strictly necessary because our `ConfigMap` client is namespace-scoped, but logs can be confusing in some cases without it.


petr-muller (approved):
/lgtm

PR #165, opened 2019-05-13
Update RPM_REPO_OPENSHIFT_ORIGIN url to be release-specific

vrutkovs:
New routes are being used to serve version-specific rpms. This also adds more tests to ensure generated URL is correct


stevekuznetsov (commented):
I'd like to see tests that reflect what we actually want to see in the jobs

stevekuznetsov:
This LGTM but I can't actually speak to the official RPM values

/assign @smarterclayton

smarterclayton:
/hold

This is not correct.  The point of `rpms....` is to provide the repo that is relative.  This change loses that.

smarterclayton:
Let me ask - why did you make this change?

vrutkovs:
Ah, I see, checked sources of artifacts-rpm container - and its a simple proxy which redirects to these urls. Closing

vrutkovs:
A better way to fix that is https://github.com/openshift/release/pull/3777

PR #167, opened 2019-05-15
rehearse: use git to identify changed templates

bbguimaraes:


pkg/config/template.go
```diff
@@ -225,28 +198,8 @@ func GetTemplateCMName(name string) string {
...
-		hash.Write(s)
-	}
-
-	// Object names can't be too long so we truncate
-	// the hash. This increases chances of collision
-	// but we can tolerate it as our input space is
-	// tiny.
-	return oneWayNameEncoding.EncodeToString(hash.Sum(nil)[:5])
+func GetTempCMName(templateName, hash string) string {
+	return fmt.Sprintf("rehearse-%s-%s", hash[:8], templateName)
```
stevekuznetsov, line 202: Can this segfault?

stevekuznetsov (commented):
Seems reasonable

/assign @petr-muller

pkg/config/template.go
bbguimaraes: First, I don't like this as a free function with seemingly disjoint parameters, which is why in the following PR it becomes a struct method: https://github.com/openshift/ci-operator-prowgen/pull/168/commits/3bc34181ab976d3c4b9c82a4fc62b1f29a8b74e9#diff-33dd159268780d36aa2b7bd9253f7a29L42.

We should not have that problem with the data flow we have. `hash` comes directly from the `git` command output, which has a well-defined format. We will either not have a template or have one with file name and hash. Someone could still call this with an inappropriate value (or an initialized struct, in #168), but a segfault is the right outcome in that situation IMO.

petr-muller:
> /assign @petr-muller

Hope to get on it today (but unlikely, meetings in the afternoon), tomorrow latest.

bbguimaraes:
Made a small change to ignore files that don't have a `.yaml` extension instead of using a blacklist:

https://github.com/openshift/ci-operator-prowgen/blob/c7634973a3ebac034518e73aebdd625c0d96db17/pkg/config/release.go#L153-L155

petr-muller (approved):
/lgtm
/approve

PR #170, opened 2019-05-29
Use upstream default trigger regexes for jobs

petr-muller:
Stop inventing our own trigger regexes for jobs, start using default
upstream ones. It looks like the upstream default regexes should:

1. Handle the CRLF newlines in bodies (our old `(?m)^/test (?:.*? )?%s(?: .*?)?$` didn't)
2. Accept comments more liberally, like `/test images, unit`


petr-muller:
Breaking change, needs https://github.com/openshift/release/pull/3907 as an immediate followup.

PR #174, opened 2019-06-11
don't pickup template jobs that will be already rehearsed

droslean:
In case of a changed template, the rehearse tool will pickup some jobs for each cluster type and template file. In that case we don't want to pick a new job with duplicated cluster type and template file with an already changed presubmit.


droslean:
@bbguimaraes @petr-muller @stevekuznetsov

petr-muller:
I don't think this is even necessary... If `AddRandomJobsForChangedTemplates` selects a job that is already being selected for rehearse because of a different reason, the duplication is removed in `AddAll` anyway, the job would be rehearsed just once.

droslean:
`AddAll()` won't remove it. The jobs can be different but they can use the same template file and cluster type.

petr-muller:
Ah! I misread the PR, sorry!

petr-muller (approved):
/lgtm

PR #175, opened 2019-06-12
[ WIP ] Introduce periodics to the rehearse tool

droslean:
This PR is focusing on including the changed periodics in the rehearse tool as well.

WIP -> Add the periodic jobs cases in Tests.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -265,6 +265,8 @@ func rehearseMain() int {
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
+	changedPeriodics := diffs.GetChangedPeriodics(masterConfig.Prow, prConfig.Prow, logger)
+
 	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
 	metrics.RecordChangedPresubmits(toRehearse)
 	metrics.RecordOpportunity(toRehearse, "direct-change")
@@ -281,9 +283,10 @@ func rehearseMain() int {
 	metrics.RecordOpportunity(toRehearseClusterProfiles, "cluster-profile-change")
 	toRehearse.AddAll(toRehearseClusterProfiles)
 
-	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
-	metrics.RecordActual(rehearsals)
-	if len(rehearsals) == 0 {
+	jobConfigurer := rehearse.NewJobConfigurer(toRehearse, changedPeriodics, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
+	presubmits, periodics := jobConfigurer.ConfigureRehearsalJobs()
+	metrics.RecordActual(presubmits)
+	if len(presubmits) == 0 && len(changedPeriodics) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		return 0
 	} else if len(rehearsals) > o.rehearsalLimit {
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -289,16 +289,16 @@ func rehearseMain() int {
 	if len(presubmits) == 0 && len(changedPeriodics) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		return 0
-	} else if len(rehearsals) > o.rehearsalLimit {
+	} else if len(presubmits) > o.rehearsalLimit {
 		jobCountFields := logrus.Fields{
 			"rehearsal-threshold": o.rehearsalLimit,
-			"rehearsal-jobs":      len(rehearsals),
+			"rehearsal-jobs":      len(presubmits) + len(changedPeriodics),
 		}
 		logger.WithFields(jobCountFields).Info("Would rehearse too many jobs, will not proceed")
 		return 0
 	}
 
-	executor := rehearse.NewExecutor(rehearsals, prNumber, o.releaseRepoPath, jobSpec.Refs, o.dryRun, loggers, pjclient)
+	executor := rehearse.NewExecutor(presubmits, periodics, prNumber, o.releaseRepoPath, jobSpec.Refs, o.dryRun, loggers, pjclient)
 	success, err := executor.ExecuteJobs()
 	metrics.Execution = executor.Metrics
 	if err != nil {
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -2,6 +2,7 @@ package diffs
 
 import (
 	"fmt"
+	"reflect"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


pkg/rehearse/jobs.go
```diff
@@ -138,17 +173,10 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
 			return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--git-ref' arg")
 		}
 	}
-	if len(source.Spec.Volumes) > 0 && !allowVolumes {
+	if len(spec.Volumes) > 0 && !allowVolumes {
 		return fmt.Errorf("jobs that need additional volumes mounted are not allowed")
 	}
 
-	if len(source.Branches) == 0 {
```
stevekuznetsov, line 145: If the periodic has `ExtraRefs` set, use the first item in that list as equivalent to a `Refs` field

stevekuznetsov:
Shouldn't this be opt-in?

droslean:
> Shouldn't this be opt-in?

It should, but it seems that we designed the code to handle only presubmits.

pkg/diffs/diffs.go
```diff
@@ -229,3 +230,26 @@ func GetPresubmitsForClusterProfiles(prowConfig *prowconfig.Config, profiles []c
...
 }
+
+// GetChangedPeriodics returns the changed presubmits mapped by their name.
+func GetChangedPeriodics(prowMasterConfig, prowPRConfig *prowconfig.Config, logger *logrus.Entry) []prowconfig.Periodic {
+	masterPeriodics := getPeriodicsPerName(prowMasterConfig.JobConfig.AllPeriodics())
+	prPeriodics := getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics())
+
+	var changedPeriodics []prowconfig.Periodic
+	for name, job := range prPeriodics {
+		if !reflect.DeepEqual(masterPeriodics[name], job) {
```
droslean, line 241: @stevekuznetsov @petr-muller @bbguimaraes @hongkailiu 
I forced to use this instead of equality's DeepEqual because of, guess what.... It panics because it can't process the `time.Duration` in decoration config :)... I tried also to use [reflect#Equalities.DeepDerivative](https://godoc.org/k8s.io/apimachinery/third_party/forked/golang/reflect#Equalities.DeepDerivative) instead but still panics... Any ideas?

cmd/pj-rehearse/main.go
```diff
@@ -265,6 +265,8 @@ func rehearseMain() int {
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
+	changedPeriodics := diffs.GetChangedPeriodics(masterConfig.Prow, prConfig.Prow, logger)
```
petr-muller, line 268: 1. Record metrics please
2. I don't like how the naming of `toRehearse` and `changedPeriodics` are inconsistent. It's confusing, because both have a semantics of "jobs that should be rehearsed". Either we can have `presubmitsToRehearse` and `periodicsToRehearse` or (IMO better) we could have a type and do something like

```go
toRehearse.Presubmits = ...
toRehearse.Periodics =...
```

cmd/pj-rehearse/main.go
```diff
@@ -281,21 +283,22 @@ func rehearseMain() int {
 	metrics.RecordOpportunity(toRehearseClusterProfiles, "cluster-profile-change")
 	toRehearse.AddAll(toRehearseClusterProfiles)
 
-	rehearsals := rehearse.ConfigureRehearsalJobs(toRehearse, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
-	metrics.RecordActual(rehearsals)
-	if len(rehearsals) == 0 {
+	jobConfigurer := rehearse.NewJobConfigurer(toRehearse, changedPeriodics, prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
+	presubmits, periodics := jobConfigurer.ConfigureRehearsalJobs()
```
petr-muller, line 287: I'm not a greatest fan of making the APIs this complex (it was complex already). Let's have multiple but simpler methods, each doing a single-better defined thing, something like this:

```go
jobConfigurer := rehearse.NewJobConfigurer(prConfig.CiOperator, prNumber, loggers, o.allowVolumes, changedTemplates, changedClusterProfiles)
presubmits := jobConfigurer.ConfigurePresubmitRehearsals(presubmitsToRehearse)
periodics := jobConfigurer.ConfigurePeriodicRehearsals(periodicsToRehearse)
```

cmd/pj-rehearse/main.go
```diff
@@ -281,21 +283,22 @@ func rehearseMain() int {
...
+	metrics.RecordActual(presubmits)
```
petr-muller, line 288: Metrics will need to be done properly. I think the current fields should be used for totals (all rehearsals opportunities, all rehearsals taken) and new fields should be added for how many of these were presumits, periodics etc.

cmd/pj-rehearse/main.go
```diff
@@ -281,21 +283,22 @@ func rehearseMain() int {
...
+	if len(presubmits) == 0 && len(changedPeriodics) == 0 {
 		logger.Info("no jobs to rehearse have been found")
 		return 0
-	} else if len(rehearsals) > o.rehearsalLimit {
+	} else if len(presubmits) > o.rehearsalLimit {
```
petr-muller, line 292: Does not seem right, I think we should count both job types against the threshold, so we never rehearse more than the $LIMIT jobs.

cmd/pj-rehearse/main.go
```diff
@@ -281,21 +283,22 @@ func rehearseMain() int {
...
 		jobCountFields := logrus.Fields{
 			"rehearsal-threshold": o.rehearsalLimit,
-			"rehearsal-jobs":      len(rehearsals),
+			"rehearsal-jobs":      len(presubmits) + len(changedPeriodics),
```
petr-muller, line 295: `changedPeriodics` does not seem right - those are all that were changed, but we do not necessarily rehearse them all. Should be `len(periodics)`?

pkg/rehearse/jobs.go
```diff
@@ -110,24 +110,59 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
+	rehearsal.Name = fmt.Sprintf("rehearse-%d-%s", prNumber, source.Name)
+	if rehearsal.Labels == nil {
+		rehearsal.Labels = make(map[string]string, 1)
+	}
+	rehearsal.Labels[rehearseLabel] = strconv.Itoa(prNumber)
+
+	return rehearsal, nil
+}
+
+func filterJobs(changedPresubmits map[string][]prowconfig.Presubmit, changedPeriodics []prowconfig.Periodic, allowVolumes bool, logger logrus.FieldLogger) (config.Presubmits, []prowconfig.Periodic) {
```
petr-muller, line 126: Suggest to split it to two, one for each job type. Simpler to grok and test.

petr-muller (commented):
Good start! Some comments inline. My main concern now would be the user experience, where failed periodic rehearsals would not be communicated to the users (I think right now jsut the `pj-rehearse` job would fail and you could find the exact job in the log, but it's not friendly). I think it's better to deal with this in a followup PR, but I would not fully enable the feature until that's sorted out. Until then, let's add a flag to enable rehearsing periodics and let's have it disabled until we finish the UX and fully test it. I would implement the flag simply as skipping detection of changed presubmits - that way we can easily test the changes do not affect the way how presubmits are rehearsed.

